PR #10, opened 2022-07-13
Add contract meta types

leighmcculloch:
### What
Add contract env meta types.

### Why
So that contracts can attach env metadata that will be used by the host. Note that this is not the same as more general contract metadata that we might want to attach to contracts in the future, such as the SDK version they were built with. This env meta is intended only to be metadata that is used by the environment itself.

The XDR being included in this PR should go through a CAP: https://github.com/stellar/stellar-protocol/issues/1265.

Related to https://github.com/stellar/stellar-xdr-next/pull/10


Stellar-contract-meta.x
```diff
@@ -0,0 +1,43 @@
...
+namespace stellar
+{
+
+struct SCMetaCustom
+{
+    opaque key<64>;
+    opaque value<64>;
+};
+
+struct SCMetaVersionSemVer
```
paulbellamy, line 19: Wouldn't you need to include the language here? And what if someone is using a different (custom) build of the sdk?

paulbellamy: Or is this just to scope and switch the host functions?

leighmcculloch: Ahh good point, there is not necessarily just one SDK in the future. We could make that different enum values?

PR #11, opened 2022-07-13
Add static NONE value

leighmcculloch:
### What
Add static NONE value.

### Why
For use in the host interface wherever a field could have a value or could not have a value. This is distinct from the static VOID value which should indicate that a field, such as a return value, will never have a value.

Close https://github.com/stellar/rs-stellar-contract-env/issues/105

cc @graydon @jonjove @jayz22


graydon:
I still don't see the utility of this vs. a union-with-VOID; every language I can think of with multiple concepts of "nothing here" (null vs. undefined, say) winds up regretting it / having users stumbling over it.

leighmcculloch:
Ok, I concede then, and I'll change Options to use VOID.

PR #16, opened 2022-08-08
Rename BINARY to BYTES

leighmcculloch:
### What
Rename all references to BINARY to BYTES.

### Why
Binary is a somewhat confusing term. When I see it in the docs it doesn't scream at me that this is the type for storing byte slices or byte arrays. Doing a quick survey of Ethereum, they use the term Bytes in both Solidity and Vyper. Thinking back to whenever I have explained Binary to others or in my head, I immediately reach for the term "bytes"

Related https://github.com/stellar/rs-soroban-sdk/pull/394
Close https://github.com/stellar/rs-soroban-sdk/issues/385


graydon (approved):
*Shrug* ok, I don't really feel strongly.

PR #21, opened 2022-08-24
Pick up changes incorrectly committed to rs-stellar-xdr

graydon:
Some recent XDR changes were incorrectly committed directly to rs-stellar-xdr. They should have been committed to this repo, with the rs-stellar-xdr repo updated _from_ here, so that core also sees them in the C++ side when it updates from here too.

This change just picks up those changes. I will next regenerate rs-stellar-xdr from them.

cc @jayz22 and @sisuresh

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -359,6 +359,33 @@ struct TransactionMetaV2
                                         // applied if any
 };
 
+enum ContractEventType
+{
+    SYSTEM = 0,
+    CONTRACT = 1
+};
+
+struct ContractEvent
+{
+    // We can use this to add more fields, or because it
+    // is first, to change ContractEvent into a union.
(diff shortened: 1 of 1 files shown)


jayz22:
I'm sorry for the trouble @graydon. So the correct process should be:
1. Make any changes to the xdr (`.x`) files manually here. 
2. Once committed, run `make build test` on the `rs-stellar-xdr` repo (without touching ), the changes here are picked up automatically, and regenerated into the `.rs` files. 
I have a couple of questions regarding the whole process which I left a comment on https://github.com/stellar/rs-stellar-xdr/pull/143/files#r953849443 
There should be a clear documentation on the process (sorry if I missed it), I'm happy to add it but I wanna make sure my understanding is clear first.

PR #24, opened 2022-08-26
Add adverts & demands to StellarMessage for pull mode

hidenori-shinohara:
XDR change for https://github.com/stellar/stellar-core/pull/3479

### DIFF
--- a/Stellar-overlay.x
+++ b/Stellar-overlay.x
@@ -47,11 +47,19 @@ struct Hello
     uint256 nonce;
 };
 
+// During the roll-out phrase, pull mode will be optional.
+// Therefore, we need a way to communicate with other nodes
+// that we want/don't want pull mode.
+// However, the goal is for everyone to enable it by default,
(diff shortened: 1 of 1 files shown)


MonsieurNicolas (changes requested):
changes look good - you have a merge commit in there that needs to be removed before we can merge

leighmcculloch:
> you have a merge commit in there that needs to be removed before we can merge

When you hit the green button in this repo GitHub will squash this PR into a single commit, so no need to rewrite the history of this PR before merging.

MonsieurNicolas:
> > you have a merge commit in there that needs to be removed before we can merge
> 
> When you hit the green button in this repo GitHub will squash this PR into a single commit, so no need to rewrite the history of this PR before merging.

The problem with merge commits is that they are not reviewable from the Github UI (at least it was a few years ago): it's possible that that commit contains bad merges for example that would end up in the main branch.

MonsieurNicolas (approved):

PR #32, opened 2022-09-15
Use LF line endings for XDR.

dmkozh:
This is to make hash checks consistent across Rust/C++ repos.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.x text -merge eol=lf


.gitattributes
```diff
@@ -0,0 +1 @@
+*.x text -merge eol=lf
```
leighmcculloch, line 1: What does the `-merge` do?

leighmcculloch (commented):
One question otherwise LGTM.

.gitattributes
leighmcculloch: Ah it turns off 3-way merge? (Assuming if I'm understanding [this](https://git-scm.com/docs/gitattributes#_merge) correctly.) That seems unrelated to the rest of this change. Why is it needed?

MonsieurNicolas:
Not needed as per https://github.com/stellar/stellar-core/pull/3534

PR #37, opened 2022-09-23
Updated xdr for token wrapper and source auth

sisuresh:
The first commit is required for https://github.com/stellar/rs-soroban-env/issues/446 and the second commit is required for https://github.com/stellar/stellar-core/issues/3546.


leighmcculloch:
Does the `HostFunction` enum also need expanding, or you found a different way to signal that we want to create a contract using the account rather than a ed25519?

```c
enum HostFunction
{
    HOST_FN_CALL = 0,
    HOST_FN_CREATE_CONTRACT = 1
};
```

sisuresh:
> Does the `HostFunction` enum also need expanding, or you found a different way to signal that we want to create a contract using the account rather than a ed25519?
> 
> ```c
> enum HostFunction
> {
>     HOST_FN_CALL = 0,
>     HOST_FN_CREATE_CONTRACT = 1
> };
> ```

Ah no good catch. I forgot to `git add` that change before committing.

leighmcculloch:
Can I suggest if we are going to add `HOST_FN_CREATE_CONTRACT_FROM_ACCOUNT`, that we also add the explicit `_FROM_ED25519` to the existing one so there is no ambiguity about what it does?

PR #38, opened 2022-09-23
add missing enum value

sisuresh:


Stellar-transaction.x
```diff
@@ -477,7 +477,8 @@ struct LiquidityPoolWithdrawOp
 enum HostFunction
 {
     HOST_FN_CALL = 0,
-    HOST_FN_CREATE_CONTRACT = 1
+    HOST_FN_CREATE_CONTRACT = 1,
+    HOST_FN_CREATE_CONTRACT_WITH_SOURCE = 2
```
leighmcculloch, line 481: If this is too late or inconvenient, don't worry about it, but if not, can we add an explicit `_FROM_ED25519` to the existing one so there is no ambiguity about what it does? Am I interpreting it right that the existing one only does from ed25519?

leighmcculloch: We can also create a follow up issue for this later, it isn't important to do now.

sisuresh: Good point. I think it's easier to just do it now - https://github.com/stellar/stellar-xdr-next/pull/39.

PR #48, opened 2022-10-07
Add an explicit error to `InvokeHostFunctionResult`.

dmkozh:
This is to be used for more useful error reporting for the failed contract invocations.

### DIFF
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -1659,7 +1659,7 @@ case INVOKE_HOST_FUNCTION_MALFORMED:
 case INVOKE_HOST_FUNCTION_TRAPPED:
     void;
 case INVOKE_HOST_FUNCTION_ERROR:
-    SCVal error;
+    SCStatus error;
 };
 
 /* High level Operation Result */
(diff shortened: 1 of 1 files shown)


Stellar-transaction.x
```diff
@@ -1657,6 +1658,8 @@ case INVOKE_HOST_FUNCTION_SUCCESS:
 case INVOKE_HOST_FUNCTION_MALFORMED:
 case INVOKE_HOST_FUNCTION_TRAPPED:
     void;
+case INVOKE_HOST_FUNCTION_ERROR:
+    SCVal error;
```
leighmcculloch, line 1662: What is the ScVal in this case?

leighmcculloch: The host and contracts use `Status` for errors, not SCVal.

graydon: Yeah this should be a Status.

leighmcculloch:
```suggestion
case INVOKE_HOST_FUNCTION_ERROR:
    SCStatus error;
```

graydon: I pushed a further commit that makes this change.

leighmcculloch (approved):

Stellar-transaction.x
dmkozh: Thanks, I was just not sure which type would be easier to consume here.

dmkozh:
@graydon, should we revive this? It still seems like a useful thing to have.

PR #51, opened 2022-11-07
XDR side of the contract deployment changes specified in CAP-46-02.

dmkozh:
This introduces the ledger entries to store the wasm code and also changes the host functions to support creating/installing the contracts.

See https://github.com/stellar/stellar-protocol/pull/1315 for the general overview of the changes.

### DIFF
--- a/Stellar-ledger-entries.x
+++ b/Stellar-ledger-entries.x
@@ -100,7 +100,8 @@ enum LedgerEntryType
     CLAIMABLE_BALANCE = 4,
     LIQUIDITY_POOL = 5,
     CONTRACT_DATA = 6,
-    CONFIG_SETTING = 7
+    CONTRACT_CODE = 7,
+    CONFIG_SETTING = 8
 };
 
 struct Signer
@@ -499,6 +500,13 @@ struct ContractDataEntry {
     SCVal val;
 };
 
+struct ContractCodeEntry {
+    ExtensionPoint ext;
+
+    Hash hash;
+    opaque code<SCVAL_LIMIT>;
+};
+
 enum ConfigSettingType
 {
     CONFIG_SETTING_TYPE_UINT32 = 0
@@ -560,6 +568,8 @@ struct LedgerEntry
         LiquidityPoolEntry liquidityPool;
     case CONTRACT_DATA:
         ContractDataEntry contractData;
+    case CONTRACT_CODE:
+        ContractCodeEntry contractCode;
     case CONFIG_SETTING:
         ConfigSettingEntry configSetting;
     }
@@ -622,6 +632,11 @@ case CONTRACT_DATA:
         Hash contractID;
         SCVal key;
     } contractData;
+case CONTRACT_CODE:
+    struct
+    {
+        Hash hash;
+    } contractCode;
 case CONFIG_SETTING:
     struct
     {
@@ -645,6 +660,7 @@ enum EnvelopeType
     ENVELOPE_TYPE_CONTRACT_ID_FROM_ED25519 = 8,
     ENVELOPE_TYPE_CONTRACT_ID_FROM_CONTRACT = 9,
     ENVELOPE_TYPE_CONTRACT_ID_FROM_ASSET = 10,
-    ENVELOPE_TYPE_CONTRACT_ID_FROM_SOURCE_ACCOUNT = 11
+    ENVELOPE_TYPE_CONTRACT_ID_FROM_SOURCE_ACCOUNT = 11,
+    ENVELOPE_TYPE_CREATE_CONTRACT_ARGS = 12
 };
 }
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -474,23 +474,79 @@ struct LiquidityPoolWithdrawOp
     int64 minAmountB; // minimum amount of second asset to withdraw
 };
 
-enum HostFunction
+enum HostFunctionType
 {
-    HOST_FN_INVOKE_CONTRACT = 0,
-    HOST_FN_CREATE_CONTRACT_WITH_ED25519 = 1,
-    HOST_FN_CREATE_CONTRACT_WITH_SOURCE_ACCOUNT = 2,
-    HOST_FN_CREATE_TOKEN_CONTRACT_WITH_SOURCE_ACCOUNT = 3,
-    HOST_FN_CREATE_TOKEN_CONTRACT_WITH_ASSET = 4
+    HOST_FUNCTION_TYPE_INVOKE_CONTRACT = 0,
+    HOST_FUNCTION_TYPE_CREATE_CONTRACT = 1,
+    HOST_FUNCTION_TYPE_INSTALL_CONTRACT_CODE = 2
+};
+
+enum ContractIDType
+{
+    CONTRACT_ID_FROM_SOURCE_ACCOUNT = 0,
+    CONTRACT_ID_FROM_ED25519_PUBLIC_KEY = 1,
+    CONTRACT_ID_FROM_ASSET = 2
+};
+ 
+enum ContractIDPublicKeyType
+{
+    CONTRACT_ID_PUBLIC_KEY_SOURCE_ACCOUNT = 0,
+    CONTRACT_ID_PUBLIC_KEY_ED25519 = 1
+};
+
+struct InstallContractCodeArgs
+{
+    opaque code<SCVAL_LIMIT>;
+};
+
+enum CreateContractSourceType {
+    CONTRACT_SOURCE_CODE_REF = 0,
+    CONTRACT_SOURCE_INSTALL_CODE = 1
+};
+
+union CreateContractSource switch (CreateContractSourceType type)
+{
+case CONTRACT_SOURCE_CODE_REF:
+    SCContractCode codeRef;
+case CONTRACT_SOURCE_INSTALL_CODE:
+    InstallContractCodeArgs installContractCodeArgs;
+};
+
+union ContractID switch (ContractIDType type)
+{
+case CONTRACT_ID_FROM_SOURCE_ACCOUNT:
+    uint256 salt;
+case CONTRACT_ID_FROM_ED25519_PUBLIC_KEY:
+    struct 
+    {
+        uint256 key;
+        Signature signature;
+        uint256 salt;
+    } fromEd25519PublicKey;
+case CONTRACT_ID_FROM_ASSET:
+    Asset asset;
+};
+
+struct CreateContractArgs
+{
+    ContractID contractID;
+    CreateContractSource source;
+};
+
+union HostFunction switch (HostFunctionType type)
+{
+case HOST_FUNCTION_TYPE_INVOKE_CONTRACT:
+    SCVec invokeArgs;
+case HOST_FUNCTION_TYPE_CREATE_CONTRACT:
+    CreateContractArgs createContractArgs;
+case HOST_FUNCTION_TYPE_INSTALL_CONTRACT_CODE:
+    InstallContractCodeArgs installContractCodeArgs;
 };
 
 struct InvokeHostFunctionOp
 {
     // The host function to invoke
     HostFunction function;
-
-    // Parameters to the host function
-    SCVec parameters;
-
     // The footprint for this invocation
     LedgerFootprint footprint;
 };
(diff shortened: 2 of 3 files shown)


Stellar-ledger-entries.x
```diff
@@ -499,6 +500,17 @@ struct ContractDataEntry {
...
+struct ContractCodeEntry {
+    Hash hash;
+    opaque code<SCVAL_LIMIT>;
+    union switch (int v)
+    {
+    case 0:
+        void;
+    }
+    ext;
+};
```
leighmcculloch, line 512: We should keep extension points at the front of types, and to use the predefined type for them. The reason for putting it at the front is so that we defer the decision on whether the type should be a chainable extension style, or a full body union style type. Here's an example of where we do that:
https://github.com/stellar/stellar-xdr-next/blob/48d5e17ae63bba0aa9725cd9d18d7438f44c07b1/Stellar-ledger.x#L394-L398

Iirc this was something @stanford-scs encouraged us to start doing in a recent protocol release.

Stellar-transaction.x
```diff
@@ -474,23 +474,87 @@ struct LiquidityPoolWithdrawOp
...
+{
+    CONTRACT_ID_FROM_PUBLIC_KEY = 0,
+    CONTRACT_ID_FROM_ASSET = 1
+};
+ 
+enum ContractIDPublicKeyType
+{
+    CONTRACT_ID_PUBLIC_KEY_SOURCE_ACCOUNT = 0,
+    CONTRACT_ID_PUBLIC_KEY_ED25519 = 1
+};
```
leighmcculloch, line 494: This relationship feels a bit backwards. It suggests a source account is a public key, when a source account contains a public key. Could "source account" be a `ContractIDType` variant instead? Then you don't need as much nesting. Nesting comes at a XDR size cost, and hierarchies make more assumptions about the future. It doesn't appear to offer a ton of value in this case. Am I missing something of value there?

```suggestion
enum ContractIDType
{
    CONTRACT_ID_FROM_SOURCE_ACCOUNT = 0,
    CONTRACT_ID_FROM_PUBLIC_KEY_ED25519 = 1,
    CONTRACT_ID_FROM_ASSET = 2
};
```

Stellar-transaction.x
```diff
@@ -474,23 +474,87 @@ struct LiquidityPoolWithdrawOp
...
+enum CreateContractSourceType {
+    CONTRACT_SOURCE_REF = 0,
+    CONTRACT_SOURCE_INSTALLED = 1
+};
+
+union CreateContractSource switch (CreateContractSourceType type)
+{
+case CONTRACT_SOURCE_REF:
+    SCContractCode codeRef;
+case CONTRACT_SOURCE_INSTALLED:
```
sisuresh, line 510: Should `CONTRACT_SOURCE_INSTALLED` be `CONTRACT_SOURCE_INSTALL` or something else? Installed makes it sounds like you’re using an already installed contract, which is actually the other option (`CONTRACT_SOURCE_REF`).

leighmcculloch (commented):
The structure of the new types is somewhat ambiguous as to how to use them. Some questions inline about it.

Stellar-transaction.x
dmkozh: I use nesting for the sake of avoiding duplication and to represent the logical structure of the data. Any kind of id inherited from a public key would contain salt, but the source of the public key itself might be different. If we add one more type of a public key in the future, it would just go into `ContractIDPublicKeyType` enum and will by definition have the salt.

I'm not feeling too strongly about that though and I agree the difference is pretty marginal. If that apparently doesn't improve clarity, we can remove the nesting as well.

Stellar-transaction.x
dmkozh: Renamed, please let me know if it's still unclear.

Stellar-transaction.x
```diff
@@ -474,23 +474,87 @@ struct LiquidityPoolWithdrawOp
...
+    InstallContractCodeArgs installContractCodeArgs;
+};
```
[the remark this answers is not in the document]
dmkozh: Renamed, please let me know if it's still unclear.

Stellar-transaction.x
leighmcculloch: >Any kind of id inherited from a public key would contain salt

This is an assumption. I don't think we know if it would be true for all ways we create contracts from public keys.

Stellar-transaction.x
```diff
@@ -474,23 +474,79 @@ struct LiquidityPoolWithdrawOp
...
+
+union HostFunction switch (HostFunctionType type)
+{
+case HOST_FUNCTION_TYPE_INVOKE_CONTRACT:
+    SCVec invokeArgs;
+case HOST_FUNCTION_TYPE_CREATE_CONTRACT:
+    CreateContractArgs createContractArgs;
+case HOST_FUNCTION_TYPE_INSTALL_CONTRACT_CODE:
+    InstallContractCodeArgs installContractCodeArgs;
```
leighmcculloch, line 543: It feels like an unnecessary optimization to support two ways to install contracts. Developers are going to want to be told one way to do it for 99% of the time. In our docs, we will want to describe one way and not have to tell a long story about when to use either way.

If installing separately is a critical feature we need, could we make that the only way to upload code? Two ops are a small cost, so we should instead make it trivial to do two ops, one install, one create.

If the ref is just a sha256, that's a small lift to require developers to provide in the second operation.

leighmcculloch (commented):
The two ways to do a thing is confusing to me, and I struggle to see us actually promoting both ways in docs. See inline comment.

Stellar-transaction.x
```diff
@@ -474,23 +474,79 @@ struct LiquidityPoolWithdrawOp
...
-enum HostFunction
+enum HostFunctionType
 {
-    HOST_FN_INVOKE_CONTRACT = 0,
-    HOST_FN_CREATE_CONTRACT_WITH_ED25519 = 1,
-    HOST_FN_CREATE_CONTRACT_WITH_SOURCE_ACCOUNT = 2,
-    HOST_FN_CREATE_TOKEN_CONTRACT_WITH_SOURCE_ACCOUNT = 3,
-    HOST_FN_CREATE_TOKEN_CONTRACT_WITH_ASSET = 4
+    HOST_FUNCTION_TYPE_INVOKE_CONTRACT = 0,
+    HOST_FUNCTION_TYPE_CREATE_CONTRACT = 1,
```
sisuresh, line 480: If I understand this correctly, we can call these three "host functions" from `InvokeHostFunctionOp`, but contracts won't necessarily have the same access. For example, it doesn't sound like contracts have access to `HOST_FUNCTION_TYPE_CREATE_CONTRACT`. Does the naming here make sense then? Should `HOST_FUNCTION` still be used here?

leighmcculloch: The invoke-host-function-invoke-contract-function nested concept is not particularly fun to explain to folks either in person, or in docs.

I don't feel strongly about where this goes, but I suspect from a developer experience point-of-view if we are adding XDR types – which is attractive – we could considering having separate operations for each Soroban operation, which falls in line with historically what we've done, and has some benefits for ingesting applications that categorize based on operation type (this isn't something that needs to drive any decision, just noting alignment).

Stellar-transaction.x
leighmcculloch: I defer to @sisuresh.

But if we are concerned about multi-op transactions, then it feels like the contract lifecycle story as a whole has some holes, because last I remember we're planning to rely on multi-op for atomic initialization. Or is that still undecided?

sisuresh: I believe the potential risk with multi-op transactions is that the fee model won't work or will become more difficult to get right. For this reason, we should be careful about making decisions off of this. We may need to add functionality for atomic initializations if the multi-op experiment doesn't work out.

But if we do want a single way to install a contract, why do both operations need to be in the same transaction? The install and create need to run sequentially, but not atomically, so you could use two separate transactions instead right?

sisuresh: This may actual expose my lack of knowledge with Horizon - Is it easy to "batch" submit two transactions? Or is the typical flow `submit tx1 -> ack tx1 in ledger -> submit tx2`?

Stellar-transaction.x
sisuresh: We could just remove the `HOST_FUNCTION_TYPE*`. You've already done a good job of abstracting away the host function details by no longer requiring the user to pass the exact `ScVec` parameters that get passed to the host function. So why even keep the concept of host functions here?

Separate operations is an interesting idea, and one that would be easy to do. One of the advantages of a single op with different options is that we don't need to plumb each new operation through core as we iterate. This is not that big of a deal, so I'm not opposed to separate operations if we see the benefit.

Stellar-transaction.x
leighmcculloch: Horizon should I think deliver both transactions into the same ledger if it can, however Horizon will synchronously block the HTTP client that is giving the transaction to Horizon, and so in practice what happens is a downstream effect where applications using Horizon only submit one transaction per ledger, because they want for that synchronous action to complete before submitting the next transaction. (cc @bartekn @tamirms to confirm if this is correct)

However, regardless of Horizon, the Soroban RPC will not synchronously block like this according to its design docs (https://stellarorg.atlassian.net/wiki/spaces/PJC/pages/4402151440/soroban-rpc+API+Docs#sendTransaction), and so applications using the RPC should be able to do as you suggest.

dmkozh: > But if we do want a single way to install a contract, why do both operations need to be in the same transaction?

I guess that's mostly a UX thing here.

> Horizon should I think deliver both transactions into the same ledger if it can,

It can do this now because we perform a significant effort to allow multiple txs from the same account per ledger. That's pretty hard to maintain and given that eventually we may want to execute soroban txs in parallel, we may not even start supporting that behavior for soroban txs even in V1. That's still TBD, but definitely something to keep in mind.

Given the feedback, I'll remove this for now. Returning both atomic installation and initialization will be a simple incremental change if we need it.

Stellar-transaction.x
dmkozh: Separate ops is something to consider before V1, but that's a relatively significant change from the core standpoint (in terms of diffs etc, not necessarily in terms of complexity). I would propose to not try doing that within this PR to keep the reviews focused on one thing.

sisuresh: I agree - separate ops should not be part of this PR and require further discussion.

dmkozh: I would actually suggest to also do the renaming separately for the same reason (unless we only rename the enum; but then the op name itself will be confusing). There is a significant amount of diffs in core/host related only to the deployment changes and I don't want to additionally mix it with no-op renaming.

Stellar-transaction.x
leighmcculloch: Hmm, this is helpful perspectives. I think separate operations for install and create only work if we can do them in a single tx from a DX point-of-view. Maybe the unnecessary piece is the ability to install without creating, even though that sounds more logically pure.

> Given the feedback, I'll remove this for now.

So it's clear, which piece is being removed?

dmkozh: I've removed the fused install option for now. I agree we should come back to it if we can't make multi-op work (the same applies to the initializers).

While install without create doesn't seem to be needed in the most of the cases, it supports the contract factories in a pretty clean fashion: you would need to install the source(s) to instantiate with the factory, install the source of the factory, then instantiate the factory contract (preferably all in the same tx if we can make it). So if we don't have a separate, then the developer would need to instantiate some throw-away contracts. I don't have intuition on usefulness of the factories (given that a non-empty set of their use-cases is already covered by decoupling the source from instance), but since we can easily come up with a use case, it doesn't feel right to make it awkward.

dmkozh: FWIW I wouldn't feel too bad if we had two options and promoted the fused option for the general use case, but have a separate documentation on 'contract factories' that refers to install specifically.

leighmcculloch: Ok, yeah this sounds fine and we can iterate on this again if the multiop doesn't work out.

Stellar-transaction.x
sisuresh: Makes sense to me.

dmkozh:
Since we seem to have agreed to keep the naming for now and do the renaming separately, is there anything else that needs to be addressed before merge?

sisuresh (approved):
@dmkozh LGTM

PR #52, opened 2022-11-17
Make events a list of lists so we can associate events with their operation

paulbellamy:
With the old schema, it is impossible to tell which operations emitted which events. This is not an issue now, as there is only one operation in soroban txns, but if we support multiple in future, it would be an issue. To prevent that we can change the events field to be a list of lists (one list per operation).

Two questions:
- Anywhere else we need to update this? (aside from propagating it through core/soroban-cli/go/js/etc...)
- Do we need the OperationEvents struct? Is there a simpler way to do an inline nested list?

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -419,8 +419,8 @@ struct TransactionMetaV3
     OperationMeta operations<>;         // meta for each operation
     LedgerEntryChanges txChangesAfter;  // tx level changes after operations are
                                         // applied if any
-    ContractEvent events<>;            // custom events populated by the
-                                        // contracts themselves
+    ContractEvent events<><>;           // custom events populated by the
+                                        // contracts themselves. One list per operation.
     TransactionResult txResult;
 
     Hash hashes[3];                     // stores sha256(txChangesBefore, operations, txChangesAfter),
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -412,6 +412,11 @@ struct ContractEvent
     body;
 };
 
+struct OperationEvents
+{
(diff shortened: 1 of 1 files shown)


Stellar-ledger.x
```diff
@@ -412,14 +412,19 @@ struct ContractEvent
...
+struct OperationEvents
+{
+    ContractEvent events<>;
+};
+
 struct TransactionMetaV3
 {
     LedgerEntryChanges txChangesBefore; // tx level changes before operations
                                         // are applied if any
     OperationMeta operations<>;         // meta for each operation
```
leighmcculloch, line 424: @sisuresh What's the rationale for keeping the events out of operation meta? I thought I heard you say there's a benefit to keeping it at the tx level, but why is unclear to me.

sisuresh: The rationale was that we'd like to hash events separately from OperationMeta (which includes the ledger entry changes for the operation). If a user wants to verify the authenticity of an event, they would hash all `events` in the transaction, combine that with the `sha256(txChangesBefore, operations, txChangesAfter)` and `sha256(txResult)`, which can both just be pulled instead of calculated, and then verify the hash of the combined hashes. 

If events were in OperationMeta, `hashes` [here](https://github.com/stellar/stellar-xdr/blob/1a76201f837e5b0f4fa5aff228ee865bfc1f6361/Stellar-ledger.x#L426) would contain `sha256(txChangesBefore, txChangesAfter)`, `sha256(operations)`, and `sha256(txResult)`. This would mean that the user would also need to hash all ledger entry changes in operations to verify an event.

PR #56, opened 2022-11-18
Generalize configuration-related XDR.

dmkozh:
The changes here reflect the recent updates to [CAP-0046-09](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-09.md):

- Use unique keys to identify settings
- Only distribute a hash for config upgrades in `StellarValue`
- Add SCP messages for exchanging config upgrade sets

### DIFF
--- a/Stellar-ledger-entries.x
+++ b/Stellar-ledger-entries.x
@@ -520,20 +520,13 @@ case CONFIG_SETTING_TYPE_UINT32:
 
 enum ConfigSettingID
 {
-    CONFIG_SETTING_CONTRACT_MAX_SIZE = 0
+    CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES = 0
 };
 
-struct ConfigSettingEntry
+union ConfigSettingEntry switch (ConfigSettingID configSettingID)
 {
-    union switch (int v)
-    {
-    case 0:
-        void;
-    }
-    ext;
-
-    ConfigSettingID configSettingID;
-    ConfigSetting setting;
+case CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES:
+    uint32 contractMaxSizeBytes;
 };
 
 struct LedgerEntryExtensionV1
--- a/Stellar-overlay.x
+++ b/Stellar-overlay.x
@@ -83,7 +83,7 @@ struct PeerAddress
     uint32 numFailures;
 };
 
-// Next ID: 18
+// Next ID: 20
 enum MessageType
 {
     ERROR_MSG = 0,
@@ -113,7 +113,11 @@ enum MessageType
 
     SEND_MORE = 16,
     FLOOD_ADVERT = 18,
-    FLOOD_DEMAND = 19
+    FLOOD_DEMAND = 19,
+    
+    // Configuration upgrades
+    GET_CONFIG_UPGRADE_SET = 20,
+    CONFIG_UPGRADE_SET = 21
 };
 
 struct DontHave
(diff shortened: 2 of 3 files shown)


Stellar-overlay.x
```diff
@@ -263,6 +267,11 @@ case SURVEY_REQUEST:
 case SURVEY_RESPONSE:
     SignedSurveyResponseMessage signedSurveyResponseMessage;
 
+case GET_CONFIG_UPGRADE_SET:
+    uint256 configUgradeSetHash;
```
sisuresh, line 271: Nit: typo `configUgradeSetHash` -> `configUpgradeSetHash`

Stellar-overlay.x
```diff
@@ -113,7 +113,11 @@ enum MessageType
 
     SEND_MORE = 16,
     FLOOD_ADVERT = 18,
-    FLOOD_DEMAND = 19
+    FLOOD_DEMAND = 19,
+    
+    // Configuration upgrades
+    GET_CONFIG_UPGRADE_SET = 20,
```
sisuresh, line 119: I have a high level question about the [CAP](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-09.md). Would it be possible to force validators to have the `ConfigUpgradeSets` locally, and avoid having to make these overlay changes? Validators would call the `upgrade` command with the full xdr, save it, but then vote on the hash. I'm still working out if this can be used maliciously.

dmkozh: That's an interesting idea. My only concern is that if we ever allow fuzzy upgrade matching, then this won't work. E.g. if a validator wants to use its own calibration data and computes a coefficient a 1000, while another one computers a coefficient of 1001. I don't know if that's ever a good idea though.

sisuresh: Yeah I don't see how adding this fuzzy matching complexity is a good idea. Is that flexibility helpful? Which setting would be applied to the network (1000 vs 1001 in your example)?. Upgrade values are meant to be determined offline. Once validator operators agree on what the value should be, they submit the settings to the network.

dmkozh: The consensus could be done on e.g. the smallest hash. The reason would be so that validators could run the calibration themselves. Otherwise there needs to be some centralized distribution of the calibration data. That's probably fine in practice, but I'm not sure if we should eliminate this possibility. Maybe it's ok to leave it out at least for v1...

sisuresh: Is there a CAP where I can read about this "calibration"? I don't think I have a good understanding of what you're referring to, and adding support for this in the overlay and consensus seems like overkill for what you get.

Stellar-ledger-entries.x
```diff
@@ -520,20 +520,13 @@ case CONFIG_SETTING_TYPE_UINT32:
```
jayz22, line 520: I think `ConfigSettingType` and `ConfigSetting` can be removed as they are not being used any more.

dmkozh: Not sure I understand; why are these not being used?

jayz22: These lines https://github.com/stellar/stellar-xdr/blob/1e5a8cff54eb70cb5002aa44f2e2bcef4a468fa1/Stellar-ledger-entries.x#L510-L519
 are outdated.

dmkozh: I'm pretty sure they're deleted, maybe you are looking at some outdated version of the file (FWIW I see your comment attached to the initial version and not the changed one)

sisuresh:
Relevant work from this PR was merged in https://github.com/stellar/stellar-xdr/pull/75.

PR #60, opened 2022-12-13
Auth concept v1

dmkozh:


Stellar-contract.x
```diff
@@ -275,5 +335,11 @@ case SCO_CONTRACT_CODE:
     SCContractCode contractCode;
 case SCO_ACCOUNT_ID:
     AccountID accountID;
+case SCO_ACCOUNT:
+    SCAccount account;
+case SCO_ADDRESS:
+    SCAddress address;
+case SCO_NONCE_KEY:
+    SCAddress nonceAddress;
```
leighmcculloch, line 343: Why does an address when used as a nonce key need its own type? I would think ScAddress could be used for this without needing to introduce a new object type.

Stellar-contract.x
```diff
@@ -255,6 +258,63 @@ struct Int128Parts {
...
+{
+    Hash contractID;
+    SCSymbol functionName;
+};
+
+struct AuthorizedInvocation
+{
+    ContractInvocation callStack<>;
+    SCVec topArgs;
+    uint64* nonce;
```
leighmcculloch, line 271: What's the significance of the `nonce` being optional?

Stellar-contract.x
```diff
@@ -255,6 +258,63 @@ struct Int128Parts {
...
+struct ContractInvocation
+{
+    Hash contractID;
+    SCSymbol functionName;
+};
+
+struct AuthorizedInvocation
+{
+    ContractInvocation callStack<>;
+    SCVec topArgs;
```
leighmcculloch, line 270: What's the significance of `top`? Is the idea that this contains all args, not just the args for a specific invocation?

Stellar-contract.x
```diff
@@ -255,6 +258,63 @@ struct Int128Parts {
...
+    uint64* nonce;
+};
+
+enum SCAccountIdType
+{
+    SC_ACCOUNT_ID_TYPE_BUILTIN_INVOKER = 0,
+    SC_ACCOUNT_ID_TYPE_BUILTIN_CLASSIC_ACCOUNT = 1,
+    SC_ACCOUNT_ID_TYPE_BUILTIN_ED25519 = 2,
+    SC_ACCOUNT_ID_TYPE_GENERIC_ACCOUNT = 3
```
leighmcculloch, line 279: What's the significance of generic account vs classic account?

leighmcculloch: It sounds like these are:
- classic account = stellar account
- generic account = contract?

Stellar-contract.x
```diff
@@ -255,6 +258,63 @@ struct Int128Parts {
...
+    SC_ADDRESS_TYPE_ED25519 = 1,
+    SC_ADDRESS_TYPE_CONTRACT = 2
+};
+
+union SCAddress switch (SCAddressType type)
+{
+case SC_ADDRESS_TYPE_CLASSIC_ACCOUNT:
+    AccountID accountId;
+case SC_ADDRESS_TYPE_ED25519:
+    Hash ed25519PublicKey;
```
leighmcculloch, line 313: Why is ed25519 public key a `Hash` and not a `PublicKey`?

Stellar-contract.x
dmkozh: This iteration has nonces managed by host and hence we need a special object type to both make it simple for host to work with it and also to not allow contracts to interact with it (similar to `ScContractCode` objects).

Stellar-contract.x
dmkozh: `Vec<AuthorizedInvocation>` is basically a flattened forest of invocation trees and only the root nodes need the nonce to be consumed in order to prevent replay. This is kind of an optimization, but I think it's also useful from the UX standpoint (e.g. it allows to do multiple ops on the same token in arbitrary order as long as the token is being used by the different contracts).

Stellar-contract.x
leighmcculloch: Since this is a recurring and rather generic problem, maybe we need an ScObject type `Internal` that contains an ScVal that serves this purpose for all types we don't want to be accessible by a contract?

Stellar-contract.x
dmkozh: These are args for the top-of-the-call-stack contract invocation. Maybe it could be named more clearly.

Stellar-contract.x
dmkozh: That's right (to be more precise, generic account = smart wallet contract with a specific interface), and from the general contract perspective everything is just an 'account', hence I wanted to make it clear that none of these entities corresponds to just 'Account'.

Stellar-contract.x
leighmcculloch: I don't think it's the name of this specific field, but rather I'm still piecing together how this works as a whole.

Stellar-contract.x
dmkozh: No reason really, probably this indeed should be `PublicKey`. As I've mentioned in the doc, we probably should drop this anyway.

Stellar-contract.x
dmkozh: That wouldn't serve the second purpose (make it easy for host to interact with this object). It would need to have some map type in order to avoid collisions with another `Internal` objects, which is both less convenient and less efficient. I would be fine with grouping all the internal objects together though, so that they don't pollute the public interface much Maybe introduce `ScInternal` `ScVal` variant to not even go to the objects? These aren't even meant to be exposed to SDKs.

Stellar-transaction.x
```diff
@@ -653,7 +653,13 @@ case ENVELOPE_TYPE_CREATE_CONTRACT_ARGS:
         Hash networkID;
         SCContractCode source;
         uint256 salt;
-    } createContractArgs;        
+    } createContractArgs;
+case ENVELOPE_TYPE_CONTRACT_AUTH:
+    struct
+    {
+        string networkPassphrase<>;
```
MonsieurNicolas, line 660: note that this should be `Hash networkID;` (which happens to be `H(networkPassphrase)`)

Stellar-contract.x
```diff
@@ -255,6 +258,63 @@ struct Int128Parts {
...
+};
+
+enum SCAddressType
+{
+    SC_ADDRESS_TYPE_CLASSIC_ACCOUNT = 0,
+    SC_ADDRESS_TYPE_ED25519 = 1,
+    SC_ADDRESS_TYPE_CONTRACT = 2
+};
+
+union SCAddress switch (SCAddressType type)
```
tamirms, line 308: it seems like `SCAddress` represents the entity which is authorizing the the invocation whereas `ScAccountId` is kind of describing how the authorized invocation should be verified? is that correct?

dmkozh: Yeah, but I'm getting rid of this in favor of having `ScAddress` everywhere. The sole purpose of `ScAccountId` was the invoker support, but it can be supported differently.

PR #61, opened 2023-01-11
Revamp of smart contract transaction XDR

dmkozh:
- Remove the mentions of 'host functions' as with the structured arguments this no longer resembles the host functions that the contract would invoke. Instead I've tried more direct 'SmartContract' name.
- Rename some of the functionality in the light of upcoming naming unification pass
- Simplify the upload (was 'install') Wasm operation


Stellar-transaction.x
```diff
@@ -66,7 +66,7 @@ enum OperationType
     SET_TRUST_LINE_FLAGS = 21,
     LIQUIDITY_POOL_DEPOSIT = 22,
     LIQUIDITY_POOL_WITHDRAW = 23,
-    INVOKE_HOST_FUNCTION = 24
+    SMART_CONTRACT = 24
```
sisuresh, line 69: Just a thought - `SMART_CONTRACT` doesn't sound like an operation, but `INVOKE_SMART_CONTRACT` isn't accurate either because this operation can be used to deploy wasm and create contracts. Most of our operations sound like an action, so it would be nice to follow that here, but I don't have a good suggestion... What do you think about something like `CALL_SMART` or `USE_SMART`?

Stellar-transaction.x
```diff
@@ -514,27 +509,27 @@ case CONTRACT_ID_FROM_ASSET:
...
-case HOST_FUNCTION_TYPE_INVOKE_CONTRACT:
-    SCVec invokeArgs;
-case HOST_FUNCTION_TYPE_CREATE_CONTRACT:
-    CreateContractArgs createContractArgs;
-case HOST_FUNCTION_TYPE_INSTALL_CONTRACT_CODE:
-    InstallContractCodeArgs installContractCodeArgs;
+case SMART_CONTRACT_OPERATION_CALL_CONTRACT:
+    SCVec callArgs;
+case SMART_CONTRACT_OPERATION_INSTANTIATE_CONTRACT:
+    InstantiateContractArgs instantiateContractArgs;
```
sisuresh, line 523: I like upload and instantiate over install and create. It's much easier to infer the order (create sounds like it should be called before install, which isn't true).

Stellar-transaction.x
dmkozh: Yeah, I'm uncertain about this as well, which is why I wanted to gather some feedback/ideas.
The tricky part is this operation may do multiple relatively different things, while all the other ops are single purpose. My idea here was that this is just the first part of the action and the second part belongs to the op itself (e.g. `SMART_CONTRACT` `INSTANTIATE` or `SMART_CONTRACT` `CALL` etc.). `USE_SMART` is probably better as a single action, but doesn't compose well. I guess the most clear option would be to just go with a separate operation for every one of these. This would require some refactoring, but it seems quite simple. WDYT?

sisuresh: This separate vs single operation point has come up before (I think @leighmcculloch asked the same question). I like how there's only a single classic operation that you need to know about to use Soroban, and it makes Soroban related checks in core simpler. For example, when we add multi-op soroban transactions, we only have to check that they're all `SMART_CONTRACT` instead of checking for a list of smart operations.

I don't feel too strongly about this though because none of our other operations work this way, but I feel like in this case a single operation makes more sense since conceptually the operation is quite different. I would like to get your thoughts on this though.

leighmcculloch: A downside of this nesting is we're adding more layers of enums that make the transactions larger. On its own I wouldn't make a decisions about that, but it also makes the XDR more complex to construct for a developer with additional layers.

For example, instead of a SMART_CONTRACT operation with CALL, INSTANTIATE, and UPLOAD sub-types, we could just make those sub-types the top-level operations:

- UPLOAD_WASM
- INSTANTIATE_CONTRACT
- CALL_CONTRACT

wildework: In my opinion, the client-side developer, I.e. someone that is using a JavaScript SDK, shouldn't even know what XDR is or how to use it, so there shouldn't be any reason to shy away from nested enums.

For example, the SDK I'm building for the Remote tool looks like this.

```javascript
const account = Account.fromKey(props.sourceAccount);
const transaction = await new Transaction()
  .sourceAccount(account)
  .operation(
    new Operation.InvokeContract()
      .contractId('xyz')
      .functionName('store')
      .functionArguments(['two', 'symbols'])
  )
  .serializedFor(soroban.network);

await transaction.signedBy(account);

const response = await soroban.network.sendTransaction(
  transaction
);
```

The only API surface that relates to the underlying XDR is `Operation.InvokeContract`. But even then, the goal is to provide an interface like this.

```javascript
const contract = new Contract()
  .id('xyz')
  .network(soroban.network);
const transaction = contract.store(['two', 'symbols']);
await transaction.requestSignature();
await soroban.network.sendTransaction(transaction);
```

Sorry for the digression... Going back to the original question, why not have `USE_SOROBAN` as the umbrella operation, and then sub-types.

- `UPLOAD_EXECUTABLE`
- `INSTANTIATE_CONTRACT`
- `CALL_CONTRACT`

tomerweller: @leighmcculloch's proposal seems to be the simplest. @dmkozh is there a reason not to go with three top level operations?

dmkozh: The reason would be if we abstain from the multi-op approach in favor of a single op with multiple host fn calls (there is still discussion about this on discord).

PR #65, opened 2023-01-21
XDR support for Auth Next

dmkozh:
- Introduce the `ScAddress` type to generically represent the address in contracts
- Introduce structured authorization data and structured signature payload to use for auth in contracts
- Introduce a special object type for the host-managed nonces


Stellar-transaction.x
```diff
@@ -653,7 +676,13 @@ case ENVELOPE_TYPE_CREATE_CONTRACT_ARGS:
         Hash networkID;
         SCContractCode source;
         uint256 salt;
-    } createContractArgs;        
+    } createContractArgs;
+case ENVELOPE_TYPE_CONTRACT_AUTH:
+    struct
+    {
+        string networkPassphrase<>;
```
tsachiherman: wouldn't networkID suffice here ?

paulbellamy: ~network passphrase is basically the stellar equivalent of the network id, no?~

dmkozh: Ok, let's use network id here. I've just got an impression that passphrase is a bit more usable, but probably this doesn't matter much. FWIW this whole thing gets hashed before signing.

tsachiherman: It's really required only when you have more than one signer, right ? ( i.e. otherwise, it's part of the transaction rather then operation )/

tsachiherman: just wondering : is there such a thing like "optional field" in XDR ?
i.e. from security perspective, it sounds like it would be needed only for the case of multiple signatures or Invokes which won't be using the transaction signature.

dmkozh: > It's really required only when you have more than one signer, right ?

No, the hash of this payload is what needs to be signed in every case besides invoker.

> just wondering : is there such a thing like "optional field" in XDR ?

There are optional fields (syntax is * as in C pointers). To be clear, this whole XDR message is *not* part of the transaction and is not for transfer across the network. This is the preimage of the signature payload (i.e. the account needs to sign sha256(preimage) and attach the signature(s) to the transaction).

tsachiherman: I see. thanks!

Stellar-transaction.x
```diff
@@ -530,12 +530,31 @@ case HOST_FUNCTION_TYPE_INSTALL_CONTRACT_CODE:
     InstallContractCodeArgs installContractCodeArgs;
 };
 
+struct AuthorizedInvocation
+{
+    Hash contractID;
+    SCSymbol functionName;
+    SCVec args;
+    AuthorizedInvocation subInvocations<>;
```
paulbellamy, line 538: Seems like with this change we're now storing the args for every invocation in the whole tree. Does that change behavior? Seems like it makes it more strict?

paulbellamy:
To clarify the tree structure. If you wanted to call a subInvocation multiple times (e.g. calling token.xfer 5 times). Would there need to be 5 children? One for each invocation?

Stellar-transaction.x
paulbellamy: It seems like this makes it harder for wallets to differentiate authorizedInvocations from incidental invocations?

dmkozh: It does make it more strict for some edge cases (which is why I switched to this structure), but overall doesn't change much. Before if you had a call stack with authorized calls `A(a, b, c)->B(d, e, f)` you'd had to sign a vector of two call stacks: `[{A(a,b,c)}, {A->B(d, e, f)}]`. Now you just sign both in a tree, so both are equivalent. The edge cases come up when the same address signs multiple stacks and it could be possible to authorize unrelated calls, which may or may not be abusable. Enforcing the tree structure makes the rules much more consistent.
FWIW in both cases the 'tree' only corresponds to the authorized calls. Calls that don't need authorization may be inserted in between, so that things like proxies/routers etc. are possible without leaking into signatures.

dmkozh:
> To clarify the tree structure. If you wanted to call a subInvocation multiple times (e.g. calling token.xfer 5 times). Would there need to be 5 children? One for each invocation?

Sure, you'll need 5 children if you have 5 different transfers (otherwise it would be possible to infinitely withdraw money from you given just a single signature). If you have many outgoing transfers and/or non-deterministic number of recipients and/or non-deterministic amounts of transfers, you should use allowance and xfer_from (we may update the token interface a bit to make these things more consistent with auth next, but that probably will happen on the next iteration).

Stellar-transaction.x
dmkozh: > It seems like this makes it harder for wallets to differentiate authorizedInvocations from incidental invocations?

What do you mean by 'incidental' invocations? As mentioned above (and as the name suggests) only authorized invocations have to be included.

The difference for wallets is walking a tree vs iterating a vector. I'd *hope* it's not a big implementation issue.

paulbellamy: By "incidental", I mean non-top-of-stack calls. For example: `A.foo -> B.bar -> C.baz`. In the case that only `C.baz` calls the verification, `A.foo`, and `B.bar` would be "incidental" because they do not call the verification. I'm making an analogy to the previous structure where only the "top-of-stack" args were included.

paulbellamy: I mean, how would a wallet select the "relevant" invocations to display to the user for approval?
Edit: to elaborate...
Previously the wallet could just display the "top-of-stack" for each stack. Which would be your token.transfers, etc.
Now, the wallet could display just the leaf nodes, but you can't tell if internal tree nodes are also authorizations (vs just incidental), so you have to show all nodes in the whole tree (which might be large, and will be harder to digest for users).

paulbellamy: We could keep the tree structure and only retain topArgs for authorizations if `args` were optional `SCVec *args;`, but if keeping all args in the tree is more secure, obviously that's better. (Just thinking out loud)

dmkozh: > In the case that only C.baz calls the verification, A.foo, and B.bar would be "incidental" because they do not call the verification. 

Again, in this case only the signature for C.baz is required. There are really no truly 'incidental' things to sign, because every call has requested a signature for some reason. I realize there can be some cases with redundancy (e.g,. consider signing something like `swap_with_some_dex->generic_swap->xfer`, the swap signature is likely redundant, but probably it would've been better if the contract writer just didn't require the `generic_swap` signature in the first place). Basically deep trees are not disallowed, but I wouldn't anticipate them to come up frequently.

In any case, I think that wallet should keep track of the relevant entities (tokens, nfts, something else) and highlight them no matter in which place in the tree they appear.

> We could keep the key structure and only retain topArgs for authorizations if args were optional SCVec *args;, but if keeping all args in the tree is more secure,

*In theory* we could omit in-between args even for authorized calls. I'm not sure if there are some attacks that are made possible by this. But if there are some, then we're in trouble, while OTOH I don't really anticipate deep trees to be prevalent, as mentioned above.

paulbellamy: I think I am misunderstanding:
> There are really no truly 'incidental' things to sign, because every call has requested a signature for some reason.

Isn't a signature only requested when a contract calls `require_auth`?

paulbellamy: > relevant entities

Ok.. I guess I was thinking of using `require_auth` as a proxy for relevance

paulbellamy: Discussed elsewhere, but my misunderstanding was that I didn't realize there were multiple trees. And the nodes in the trees are only methods which call `require_auth`. Non-authenticating methods are skipped.

Stellar-transaction.x
```diff
@@ -530,12 +530,31 @@ case HOST_FUNCTION_TYPE_INSTALL_CONTRACT_CODE:
...
+};
+
+struct ContractAuth
+{
+    SCAddress* address; // not present for invoker
+    AuthorizedInvocation invocation;
+    SCVec signatureArgs;
+    uint64* nonce; // not present for invoker
```
leighmcculloch: `ContractAuth` might be more extensible, and clearer, if we make it a union where Invoker only has two fields, and other auth forms have all four. Or a struct containing a union where the two fields are in the struct plus a union with additional fields.

Stellar-transaction.x
dmkozh: Here is more formal definition for building the tree in the signature payload:

Consider the rooted directed sub-contract call tree created by a single top-level contract invocation. From the perspective of a single address there are two kinds of nodes in the tree. Let's call them "N-nodes" - the nodes that don't require auth for the address and "A-nodes" - the nodes that call `require_auth` for that address. Now we perform the following procudure to create the signature payload: remove all the edges and all the N-nodes from the tree (so we're left with only A-nodes). Then create directed edges between the A-nodes such that there is an edge from A1 to A2 if and only if there is a path from A1 to A2 in the original tree such that only passes through N-nodes (or no other nodes at all). 

Note, that this procedure may result in multiple trees, but this is more of an edge case (that would mean that the contract makes some unrelated calls). Multiple trees would be represented by multiple `ContractAuth` entries here.

leighmcculloch, line 538: The name `subInvocations` is intriguing because to me it implies it does not contain an authorization for the top-level invocation. Is that true? I thought that would be false. Is there a reason to not call this `authInvocations` or `authorizedInvocations`?

Stellar-contract.x
```diff
@@ -255,6 +256,20 @@ struct Int128Parts {
     uint64 hi;
 };
 
+enum SCAddressType
+{
+    SC_ADDRESS_TYPE_CLASSIC_ACCOUNT = 0,
```
leighmcculloch: I think we should remove the term classic here. On Stellar an account is an account.
```suggestion
    SC_ADDRESS_TYPE_ACCOUNT = 0,
```

Stellar-contract.x
```diff
@@ -273,7 +288,9 @@ case SCO_BYTES:
     opaque bin<SCVAL_LIMIT>;
 case SCO_CONTRACT_CODE:
     SCContractCode contractCode;
-case SCO_ACCOUNT_ID:
-    AccountID accountID;
+case SCO_ADDRESS:
+    SCAddress address;
+case SCO_NONCE_KEY:
+    SCAddress nonceAddress;
```
leighmcculloch, line 294: Where can I read more about what a nonce key is, and why it contains an address? I've missed something on how this works and don't understand what it is or why it contains an address.

Stellar-transaction.x
dmkozh: I'm just still not quite sure about special-casing the invoker at all. It actually could be safer to always explicitly specify the address and maybe even use nonce (so only skipping the signature is allowed). But if we stick to special-casing invoker, I agree that maybe we should group these together.

Stellar-transaction.x
dmkozh: No special reason really, I can rename this. The naming is totally up to debate, that's just the first thing I came up with.

Stellar-contract.x
dmkozh: That makes sense. This is in artifact from the prototype where there is also 'smart' account.

Stellar-contract.x
dmkozh: I think for the first iteration we'd go with the nonce that works the same way as in the current auth (i.e. it's keyed by contract id + address). Going forward I think we have an agreement to switch to temporarily stored nonces. But that wouldn't make much sense before we have temp entry support.

Stellar-transaction.x
leighmcculloch: That's a good point 👍🏻.

There may be value in introducing a union here somewhere so that we can take a second stab at auth if we need to. Although, I guess we can add a new operation for that if we need to. So yeah, this is fine.

paulbellamy: Union here would definitely be an easier upgrade than `InvokeHostFunctionV2` though.

Stellar-transaction.x
dmkozh: Actually after re-reading your question, it seems like there is some confusion. These indeed don't contain the top-level invocation, because the represents the tree (so the contract id/fn name/args are related to the current node and subInvocations to the children). Maybe `authorizedSubInvocations` would be a bit more explicit, though it feels a bit too verbose.

paulbellamy: I guess the top-level invocation is always known, because it is in the args to the `invokeHostFunction`. In the case where the top-level invocation calls `auth_required`, I assume that would need signing, and appear as an AuthorizedInvocation?

dmkozh: In a general case args of the top-level invocation (or any invocation for that matter) != args that need to be signed (this is especially true when multiple signers are involved).

sisuresh (approved):

PR #67, opened 2023-02-03
Make unit/void case explicit and support more tuples in union spec

leighmcculloch:
### What
Make unit/void case explicit and support more tuple values in unions in the contract spec.

### Why
@brson is adding support for multiple value tuples in the contract spec. We need to change the union case spec to support multiple types so that a union case can contain more than one value.

While making that change I realized that we should probably distinguish between unit (void) and tuple variant types because an empty tuple and unit variant aren't exactly the same.

And by having a discriminant in there for them we can also add struct-like variants in the future if we choose to in a way that isn't breaking.

For https://github.com/stellar/rs-soroban-sdk/pull/850

### DIFF
--- a/Stellar-contract-spec.x
+++ b/Stellar-contract-spec.x
@@ -157,9 +157,9 @@ enum SCSpecUDTUnionCaseV0Kind
 union SCSpecUDTUnionCaseV0 switch (SCSpecUDTUnionCaseV0Kind kind)
 {
 case SC_SPEC_UDT_UNION_CASE_VOID_V0:
-    SCSpecUDTUnionCaseVoidV0 void;
+    SCSpecUDTUnionCaseVoidV0 voidCase;
 case SC_SPEC_UDT_UNION_CASE_TUPLE_V0:
-    SCSpecUDTUnionCaseTupleV0 tuple;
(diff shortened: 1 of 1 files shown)


Stellar-contract-spec.x
```diff
@@ -135,11 +135,31 @@ struct SCSpecUDTStructV0
...
+enum SCSpecUDTUnionCaseV0Kind
+{
+    SC_SPEC_UDT_UNION_CASE_VOID_V0 = 0,
+    SC_SPEC_UDT_UNION_CASE_TUPLE_V0 = 1,
+};
+
+union SCSpecUDTUnionCaseV0 switch (SCSpecUDTUnionCaseV0Kind kind)
+{
+case SC_SPEC_UDT_UNION_CASE_VOID_V0:
+    SCSpecUDTUnionCaseVoidV0 void;
```
dmkozh: `void` is prohibited by XDR spec as an identifier (https://www.rfc-editor.org/rfc/rfc4506#section-6.4). Our parsers may or may not care, but I think it's better be on the safe side here.

leighmcculloch: Good call, thanks. 3a6327d

dmkozh (approved):

PR #69, opened 2023-02-06
Add auth errors

sisuresh:


Stellar-contract.x
```diff
@@ -128,6 +129,16 @@ enum SCHostStorageErrorCode
     HOST_STORAGE_GET_ON_DELETED_KEY = 5
 };
 
+enum SCHostAuthErrorCode
+{
+    HOST_AUTH_UNKNOWN_ERROR = 0,
+    HOST_AUTH_WRONG_NONCE = 1,
```
dmkozh: Can we rename this to just `NONCE_ERROR` in order to not touch it anymore when we change the nonce approach?

Stellar-contract.x
```diff
@@ -128,6 +129,16 @@ enum SCHostStorageErrorCode
...
+    HOST_AUTH_GET_RECORDING_IN_ENFORCING_MODE = 2,
```
dmkozh: This doesn't seem right - recording mode is not a part of the protocol and it's not possible to trigger this from a transaction. So this should probably be just a general error with an appropriate message.

Stellar-contract.x
sisuresh: Done

Stellar-contract.x
sisuresh: Removed HOST_AUTH_GET_RECORDING_IN_ENFORCING_MODE

dmkozh (approved):

PR #70, opened 2023-02-11
Value representation overhaul

graydon:
This is part of https://github.com/stellar/rs-soroban-env/issues/679,  it does several things at once:

  - Remove the object/value split at the XDR level
  - Expand size of value tag from 8 to 128 cases
  - Remove `U63` (all small-number optimization is hidden from XDR but still happens in RawVal)
  - Remove `Static`, fold cases into the value cases (thus Void, Bool are just ScVal types now)
  - Remove `Bitset`, nobody was using it
  - Allow `Symbol` to be arbitrarily long, spilling into a host object as needed
  - Add `String` which is a host object like `Bytes` but displays as text
  - Add {u,i}256
  - Add special case u64 subtypes `Timepoint` and `Duration`
 
There are coupled changes in rs-stellar-xdr, rs-soroban-env and rs-soroban-sdk:

  - https://github.com/stellar/rs-stellar-xdr/pull/234
  - https://github.com/stellar/rs-soroban-env/pull/682
  - https://github.com/stellar/rs-soroban-sdk/pull/879


Stellar-contract-spec.x
```diff
@@ -17,19 +17,23 @@ enum SCSpecType
...
-    SC_SPEC_TYPE_BYTES = 11,
-    SC_SPEC_TYPE_INVOKER = 12,
-    SC_SPEC_TYPE_ADDRESS = 13,
+    SC_SPEC_TYPE_BOOL = 1,
+    SC_SPEC_TYPE_VOID = 2,
+    SC_SPEC_TYPE_STATUS = 3,
+    SC_SPEC_TYPE_U32 = 4,
+    SC_SPEC_TYPE_I32 = 5,
+    SC_SPEC_TYPE_U64 = 6,
+    SC_SPEC_TYPE_I64 = 7,
```
leighmcculloch, line 26: Now that we have so much bit space, should we also add 8 and 16-bit integers? I'm fine with no.

Stellar-contract.x
```diff
@@ -5,67 +5,73 @@
...
+    // SCContractCode and SCAddressType are types that gets used separately from
+    // SCVal so we do not flatten their structures into separate SCVal cases.
+    SCV_CONTRACT_CODE = 18,
+    SCV_ADDRESS = 19,
+
+    // SCV_LEDGER_KEY_CONTRACT_CODE and SCV_LEDGER_KEY_ADDRESS_NONCE are unique
+    // symbolic SCVals used as the key for ledger entries for a contract's code
+    // and an address' nonce, respectively.
+    SCV_LEDGER_KEY_CONTRACT_CODE = 20,
+    SCV_LEDGER_KEY_ADDRESS_NONCE = 21
```
leighmcculloch: @dmkozh Should we call this simply "Nonce"? Address nonce makes it sound like the nonce is keyed by the address, but my understanding was it is keyed by the address and contract.

Stellar-contract.x
```diff
@@ -270,27 +226,81 @@ case SC_ADDRESS_TYPE_CONTRACT:
...
-case SCO_MAP:
-    SCMap map;
-case SCO_U64:
+
+case SCV_BOOL:
+    bool b;
+case SCV_VOID:
+    void;
+case SCV_STATUS:
+    SCStatus error;
```
leighmcculloch, line 265: Can we name this variable status to reduce the concept/language space? Alternatively, rename SCStatus to SCError?
```suggestion
    SCStatus status;
```

Stellar-contract.x
```diff
@@ -270,27 +226,81 @@ case SC_ADDRESS_TYPE_CONTRACT:
...
+    int32 i32;
+
+case SCV_U64:
     uint64 u64;
-case SCO_I64:
+case SCV_I64:
     int64 i64;
-case SCO_U128:
+case SCV_TIMEPOINT:
+    uint64 timepoint;
```
leighmcculloch: Can we use the timepoint type if the goal is to align this value with that existing concept in the XDR?
```suggestion
    Timepoint timepoint;
```

Stellar-contract.x
```diff
@@ -270,27 +226,81 @@ case SC_ADDRESS_TYPE_CONTRACT:
...
+case SCV_DURATION:
+    uint64 duration;
```
leighmcculloch:
```suggestion
    Duration duration;
```

Stellar-contract.x
```diff
@@ -270,27 +226,81 @@ case SC_ADDRESS_TYPE_CONTRACT:
...
-    SCContractCode contractCode;
-case SCO_ADDRESS:
+
+case SCV_U256:
+    opaque u256[32];
+case SCV_I256:
+    opaque i256[32];
+
+case SCV_BYTES:
+    opaque bytes<SCVAL_LIMIT>;
```
leighmcculloch: I think we'd benefit from this definition being in a type like `SCString`. In the Rust XDR it will end up as a type alias.
```suggestion
    SCBytes bytes;
```

leighmcculloch (commented):
Really like the direction we're going in with this change. A few requests inline.

Stellar-contract.x
leighmcculloch: I think we'd benefit from renaming Status to Error. We use the term "Error" with developers in docs.

Stellar-contract.x
dmkozh: Yeah, +1 for just `SCV_LEDGER_KEY_NONCE`. This is also more future-proof - when/if we change the nonce structure, we won't need to update the name.

Stellar-contract.x
```diff
@@ -5,67 +5,73 @@
...
+
+    // SCContractCode and SCAddressType are types that gets used separately from
+    // SCVal so we do not flatten their structures into separate SCVal cases.
+    SCV_CONTRACT_CODE = 18,
+    SCV_ADDRESS = 19,
+
+    // SCV_LEDGER_KEY_CONTRACT_CODE and SCV_LEDGER_KEY_ADDRESS_NONCE are unique
+    // symbolic SCVals used as the key for ledger entries for a contract's code
+    // and an address' nonce, respectively.
+    SCV_LEDGER_KEY_CONTRACT_CODE = 20,
```
dmkozh: If we're doing the renaming anyway, shall we name this `SCV_LEDGER_KEY_CONTRACT_EXECUTABLE` (and `SCV_CONTRACT_CODE`->`SCV_CONTRACT_EXECUTABLE`)? I think that's the best we have now for this entity, from the discussion in https://docs.google.com/document/d/10O6cWjOGnvhH4YaVRlR0CvBygeMcbY5hdT0Tq_pzlSM

Stellar-contract.x
dmkozh: +1 for Error, I have an expectation for status to have a positive variant (like Status::Ok/Success), while IIUC we have only errors here.

Stellar-contract.x
leighmcculloch: +1

Stellar-contract.x
dmkozh: Actually I've noticed that `SCStatus` does have a positive variant. Then this shouldn't be called `error` in the first place, should it?

leighmcculloch: I don't think we use the positive form of status anywhere, and we should just focus on this being an error instead.

jayz22: `contract_event` recording returns `ScStatus::Ok` if the the event was successfully recorded. https://github.com/stellar/rs-soroban-env/blob/main/soroban-env-common/env.json#L56

leighmcculloch: It looks like in the code there's very little use of `Ok`. We could change any cases where we are returning Ok to returning Void, and get rid of Ok.

Stellar-contract-spec.x
graydon: Neither 8 nor 16 bit values has any meaning (or is even representable as a type) at the XDR or WASM level. So I think "no".

Stellar-contract.x
graydon: I actually made this change earlier and then changed it back for two reasons:

  - There are a bunch of other renamings I want to do that I thought we could do a "semantic no-op, just renaming" PR and bundle them all together.
  - This _particular_ renaming introduces a ton of churn because `Error` is a type-name used ubiquitously in Rust code already (as an associated type of lots of traits)

On the other most of the renaming I had in mind is env/sdk-side rather than XDR-side, and if you want to minimize the number of XDR PRs I'm willing to try including the Status=>Error change in this one. I totally agree it should happen eventually.

leighmcculloch: 👍 let's do it separately.

Stellar-contract.x
graydon: fixed.

Stellar-contract.x
graydon: Fixed (will push a rebased change)

Stellar-contract.x
graydon: Fixed (will push a rebased change)

Stellar-contract.x
graydon: Fixed (will push a rebased change)

Stellar-contract.x
graydon: Fixed (will push a rebased change)

Stellar-contract.x
graydon: will hold on Status => Error for now (added to tasklist on #679)

graydon:
@leighmcculloch by adding `typedef opaque SCBytes<SCVAL_LIMIT>` I wind up with a new `struct ScBytes(BytesM<256000>)` in the generated rust code, whereas `typedef string SCString<SCVAL_LIMIT>` just introduces a `type ScString = StringM<256000>`. I'm not sure what the difference is in how xdrgen treats string typedefs vs. other typedefs, but .. the additional `ScBytes(...)` wrapper is a little annoying. Is it intentional? Can I turn it off? It adds nothing here.

leighmcculloch:
That sounds like a bug. Should be a quick fix. I'll take a look today.

sisuresh (approved):

PR #74, opened 2023-03-13
Add diagnosticEvents

sisuresh:
XDR for https://github.com/stellar/rs-soroban-env/pull/713 and the WIP core change.

The env PR needs to be updated to use the new event type.

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -388,7 +388,8 @@ struct TransactionMetaV2
 enum ContractEventType
 {
     SYSTEM = 0,
-    CONTRACT = 1
+    CONTRACT = 1,
+    DIAGNOSTIC = 2
 };
 
 struct ContractEvent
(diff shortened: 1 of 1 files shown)


Stellar-ledger.x
```diff
@@ -430,6 +442,11 @@ struct TransactionMetaV3
 
     Hash hashes[3];                     // stores sha256(txChangesBefore, operations, txChangesAfter),
                                         // sha256(events), and sha256(txResult)
+
+    // Diagnostics events that are not hashed. One list per operation.
+    // This will contain all contract and diagnostic events. Even ones
+    // that were emitted in a failed contract call.
+    OperationDiagnosticEvents diagnosticEvents<>;
```
Shaptic, line 449: Does that mean this is a subset of `OperationEvents`, or they're disjoint?

sisuresh: `events<>` will be a  subset of `diagnosticEvents<>`.

Shaptic: o shoot okay so this at least doubles the size of the event-related meta, right? why can't `TransactionMetaV3.events` mark diagnostic events w/ their type accordingly? is it because you'd need to make `ContractEvent` a union to include the extraneous `inSuccessfulContractCall` field?

sisuresh: `events` will be hashed into the ledger and will be part of the protocol. We want diagnostic events to not affect the protocol, which is why we keep it separate (and duplicate regular contract events in to maintain ordering).

graydon (approved):
LGTM

PR #75, opened 2023-03-21
XDR for upgradeable config entries

sisuresh:
This contains changes from https://github.com/stellar/stellar-xdr/pull/27 and a change for upgrading config entries for https://github.com/stellar/stellar-core/pull/3674.

### DIFF
--- a/Stellar-ledger-entries.x
+++ b/Stellar-ledger-entries.x
@@ -505,17 +505,6 @@ struct ContractCodeEntry {
     opaque code<SCVAL_LIMIT>;
 };
 
-enum ConfigSettingType
-{
-    CONFIG_SETTING_TYPE_UINT32 = 0
-};
-
-union ConfigSetting switch (ConfigSettingType type)
-{
-case CONFIG_SETTING_TYPE_UINT32:
-    uint32 uint32Val;
-};
-
 enum ConfigSettingID
 {
     CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES = 0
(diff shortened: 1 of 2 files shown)


dmkozh (approved):

Stellar-ledger-entries.x
```diff
@@ -518,20 +518,13 @@ case CONFIG_SETTING_TYPE_UINT32:
```
jayz22: Is the `ConfigSetting` still needed? I see the `ConfigSettingEntry` below is using `uint32` directly so maybe this is obsolete?

dmkozh: Good catch, we indeed don't use this anymore.

sisuresh: Yeah good catch. Removed.

Stellar-ledger.x
```diff
@@ -130,6 +130,11 @@ enum LedgerUpgradeType
     LEDGER_UPGRADE_CONFIG = 6
 };
 
+struct ConfigUpgradeSetKey {
+    Hash contractID;
+    Hash contentHash;
+};
+
```
jayz22, line 137: If I'm understanding correctly, the `ConfigUpgradeSetKey` here must be a contract data belong to `ContractID`? Does that mean only a ContractData entry can be upgraded? 
I'm thinking in the context of network-wide setting such as metering cost params that wouldn't belong to a contract, how will it work?

sisuresh: The actual `ConfigSettingEntry` is not associated with a contract and is not upgraded in Soroban. We are just adding a mechanism where you can look up the xdr for a new `ConfigSettingEntry` in `ContractData`, and update the existing `ConfigSettingEntry` to use that.

PR #76, opened 2023-03-31
Add contract executable to the auth payload.

dmkozh:
Add contract executable to the auth payload.

Also did some passing-by renaming from source to executable.

### DIFF
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -517,7 +517,7 @@ case CONTRACT_ID_FROM_ASSET:
 struct CreateContractArgs
 {
     ContractID contractID;
-    SCContractExecutable source;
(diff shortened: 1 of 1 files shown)


dmkozh:
As per Soroban design discussion we have decided that this feature is likely unnecessary as on one hand it comes with a significant cost (both UX cost of transaction invalidation in case of the implementation upgrade, as well as fee cost due to transaction size and compute increase), while on the other hand doesn't provide too much protection as it's tricky to execute a presigned transaction with a new source anyway due to the ledger footprint mismatch. The latter may be reconsidered and then we might come back to this or to some in-between solution (like to allow the users to optionally add the wasm hashes to the transaction and enforce them at runtime).

PR #79, opened 2023-04-11
Add fee and limit-related configuration XDR

dmkozh:
The changes are based on [CAP-0046-07.](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-07.md)

### DIFF
--- a/Stellar-ledger-entries.x
+++ b/Stellar-ledger-entries.x
@@ -505,15 +505,112 @@ struct ContractCodeEntry {
     opaque code<SCVAL_LIMIT>;
 };
 
+// Identifiers of all the network settings.
 enum ConfigSettingID
 {
-    CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES = 0
+    CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES = 0,
+    CONFIG_SETTING_CONTRACT_COMPUTE_V0 = 1,
+    CONFIG_SETTING_CONTRACT_LEDGER_COST_V0 = 2,
+    CONFIG_SETTING_CONTRACT_HISTORICAL_DATA_V0 = 3,
+    CONFIG_SETTING_CONTRACT_META_DATA_V0 = 4,
+    CONFIG_SETTING_CONTRACT_BANDWIDTH_V0 = 5
+};
+
+// "Compute" settings for contracts (instructions and memory).
+struct ConfigSettingContractComputeV0
+{
+    // Maximum instructions per ledger
+    int64 ledgerMaxInstructions;
+    // Maximum instructions per transaction
+    int64 txMaxInstructions;
+    // Cost of 10000 instructions
+    int64 feeRatePerInstructionsIncrement;
+
+    // Memory limit per contract/host function invocation. Unlike 
+    // instructions, there is no fee for memory and it's not
+    // accumulated between operations - the same limit is applied 
+    // to every operation.
+    uint32 memoryLimit;
+};
+
+// Ledger access settings for contracts.
+struct ConfigSettingContractLedgerCostV0
+{
+    // Maximum number of ledger entry read operations per ledger
+    uint32 ledgerMaxReadLedgerEntries;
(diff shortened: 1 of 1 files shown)


graydon (approved):
I'm going to rubber-stamp this for now -- there are some minor details I can imagine discussing/debating but we're still at the roughing-out-the-basic-structure phase for fees, and I want to avoid any holdup in getting _something_ working.

dmkozh:
> there are some minor details I can imagine discussing/debating

Please let me know - for now my goal is to just introduce parameters from the CAP that we can use/tune (this is going to be a gradual process).

Stellar-ledger-entries.x
```diff
@@ -505,15 +505,112 @@ struct ContractCodeEntry {
     opaque code<SCVAL_LIMIT>;
 };
 
+// Identifiers of all the network settings.
 enum ConfigSettingID
 {
-    CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES = 0
+    CONFIG_SETTING_CONTRACT_MAX_SIZE_BYTES = 0,
```
anupsdf, line 511: I wonder if we should drop the prefix CONFIG_SETTING_ in the enum entries for better readability. These entries are within the enum ConfigSettingID so prefix feels redundant.

Stellar-ledger-entries.x
```diff
@@ -505,15 +505,112 @@ struct ContractCodeEntry {
...
+    CONFIG_SETTING_CONTRACT_HISTORICAL_DATA_V0 = 3,
+    CONFIG_SETTING_CONTRACT_META_DATA_V0 = 4,
+    CONFIG_SETTING_CONTRACT_BANDWIDTH_V0 = 5
+};
+
+// "Compute" settings for contracts (instructions and memory).
+struct ConfigSettingContractComputeV0
+{
+    // Maximum instructions per ledger
+    int64 ledgerMaxInstructions;
```
anupsdf, line 523: Should we use uint64 for this and the other two below since these will always be non-negative?

Stellar-ledger-entries.x
dmkozh: Currently XDR gets compiled into an `enum` (i.e. not into `enum class`) hence we need to 'namespace' these in such fashion.

Stellar-ledger-entries.x
dmkozh: I wouldn't worry much about representing non-negative values via unsigned values (especially in C++). But here the main thing is that the transaction XDR typically doesn't expose uint64 parameters (IIUC for better compatibility with clients such as javascript) and these values will be compared to the transaction values. I don't feel super strongly about this, but it at least seems a bit more consistent with the existing XDR definitions.

Stellar-ledger-entries.x
```diff
@@ -505,15 +505,112 @@ struct ContractCodeEntry {
...
+};
+
+// Historical data (pushed to core archives) settings for contracts.
+struct ConfigSettingContractHistoricalDataV0
+{
+    int64 feeHistorical1KB; // Fee for storing 1KB in archives
+};
+
+// Meta data (pushed to downstream systems) settings for contracts.
+struct ConfigSettingContractMetaDataV0
```
sisuresh, line 580: Is this for TransactionMeta? I don't think we refer to "meta" as "metadata" anywhere, so this might be confusing. I would just remove the "data" portion of the name.

Stellar-ledger-entries.x
leighmcculloch: >transaction XDR typically doesn't expose uint64 parameters (IIUC for better compatibility with clients such as javascript)

Fwiw `int64` also exceeds the limits of JavaScript runtimes for what can be stored in a `Number` since a `Number` can only store lossless a 53-bit number and so avoiding `uint64` in favor of `int64` doesn't address the JavaScript issue. JavaScript also has `BigInt` now which alleviates the concern.

However, I've always appreciated the use of signed integers for any length or size values because it avoids underflow bugs.

PR #81, opened 2023-04-18
Define config settings for contract costs

jayz22:


Stellar-contract-cost-params.x
```diff
@@ -0,0 +1,60 @@
+%#include "xdr/Stellar-types.h"
```
dmkozh: While I'm not a huge fan of jamming all the XDR together, this looks too granular compared to the current XDR organization. Probably it would make more sense to have all the configuration-related definitions be defined in the same file (like Stellar-network-configuration)

jayz22: Yeah I'm good with this suggestion. In particular the fees/limit, budget, cost params are all kind of related and make sense to live in the same file.

jayz22: I've split the config settings to a separate file.

graydon (approved):
LGTM

dmkozh (approved):

PR #82, opened 2023-04-18
Meta xdr

sisuresh:
Bringing back https://github.com/stellar/stellar-xdr/pull/49 for https://github.com/stellar/rs-soroban-sdk/pull/914.

### DIFF
--- a/Stellar-contract-meta.x
+++ b/Stellar-contract-meta.x
@@ -12,7 +12,7 @@ namespace stellar
 struct SCMetaV0
 {
     string key<>;
-    string val<256>;
+    string val<>;
 };
 
 enum SCMetaKind
--- a/Stellar-contract-meta.x
+++ b/Stellar-contract-meta.x
@@ -9,15 +9,21 @@
 namespace stellar
 {
 
+struct SCMetaV0
+{
+    string key<10>;
+    string val<256>;
+};
+
 enum SCMetaKind
 {
-    SC_META_KIND_SDK = 0
+    SC_META_V0 = 0
 };
 
 union SCMetaEntry switch (SCMetaKind kind)
 {
-case SC_META_KIND_SDK:
-    string sdk<256>;
+case SC_META_V0:
+    SCMetaV0 v0;
 };
 
 }
--- a/Stellar-contract-meta.x
+++ b/Stellar-contract-meta.x
@@ -1,0 +1,23 @@
+// Copyright 2022 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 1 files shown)


Stellar-contract-meta.x
```diff
@@ -0,0 +1,29 @@
+// Copyright 2022 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+// The contract spec XDR is highly experimental, incomplete, and still being
+// iterated on. Breaking changes expected.
```
leighmcculloch:
```suggestion
// The contract meta XDR is highly experimental, incomplete, and still being
// iterated on. Breaking changes expected.
```

Stellar-contract-meta.x
```diff
@@ -0,0 +1,29 @@
...
+
+% #include "xdr/Stellar-types.h"
+namespace stellar
+{
+
+struct SCMetaV0
+{
+    string key<10>;
+    string val<256>;
```
leighmcculloch: I don't remember where these size limits came from, but they're arbitrary. Maybe 1024? I guess we could even make them unbounded (which is really bounded to u32::MAX), but some reasonable limit would be better.

leighmcculloch (approved):
Couple suggestions, otherwise 👍🏻.

Stellar-contract-meta.x
sisuresh: Unbounded might make more sense to give users flexibility. The WASM contract will already be subject to a limit, and users could write whatever they want into the custom section if they wanted to.

leighmcculloch: 👍🏻

leighmcculloch (approved):

Stellar-contract-meta.x
```diff
@@ -0,0 +1,29 @@
...
+// The contract meta XDR is highly experimental, incomplete, and still being
+// iterated on. Breaking changes expected.
+
+% #include "xdr/Stellar-types.h"
+namespace stellar
+{
+
+struct SCMetaV0
+{
+    string key<10>;
```
sisuresh:
```suggestion
    string key<>;
```

Stellar-contract-meta.x
```diff
@@ -0,0 +1,29 @@
...
+// iterated on. Breaking changes expected.
+
+% #include "xdr/Stellar-types.h"
+namespace stellar
+{
+
+struct SCMetaV0
+{
+    string key<>;
+    string val<256>;
```
sisuresh:
```suggestion
    string val<>;
```

Stellar-contract-meta.x
sisuresh: Done

leighmcculloch (approved):

PR #84, opened 2023-04-24
Transaction changes to support Soroban fee model.

dmkozh:
- Moved all the resources (including the footprint) to the transaction level.
- Added host fn batching to compensate for removal of multi-operation tx support
- Did some passing-by naming cleanup (as we never really have time for that)


Stellar-transaction.x
```diff
@@ -551,17 +547,24 @@ struct ContractAuth
...
     ContractAuth auth<>;
 };
 
+struct InvokeHostFunctionOp
+{
+    // The host functions to invoke. The functions will be executed
+    // in the same fashion as operations: either all functions will
+    // be successfully applied or all fail if at least one of them
+    // fails.
+    HostFunction functions<MAX_OPS_PER_TX>;
```
sisuresh, line 565: Was this limit chosen arbitrarily? I'm wondering if it would make sense to limit it to a lower number and we can increase it in a future protocol version if we need to. I guess the tx size limit will kick in anyways so this one doesn't matter as much.

dmkozh: My only consideration was to not make the number of functions larger than max ops/tx in order to not create wrong incentives for using this vs classic ops. I'm ok with setting this to a lower number, though I'm not sure what would be the reason to do so (you probably can do 100 vm-less invocations, but with vm there are just too many limits, e.g. with the current cpu limit of 40M instructions we can have just up to 40 vm invocations).

sisuresh: Yeah the other limits should be sufficient then.

sisuresh (approved):

PR #88, opened 2023-04-28
Replace ContractAuth addressWithNonce option with explicit union

leighmcculloch:
### What
Replace ContractAuth addressWithNonce option with explicit union.

### Why
The ContractAuth addressWithNonce field is an option field, essentially a union, between no value, and some value containing an address. We currently use the no value state to implicitly communicate that the authorization is intended to be by the source account of the transaction operation.

Because this behavior is implied by the optional being unset, the behavior is not particularly obvious or intuitive. As a first-time user of the field I could not arrive at why the field was optional on my own, and later when attemting to figure out how to specify that an auth would be by a source account, I could not figure out which field to control that with. This is particularly confusing since source accounts can be expressed as addresses.

Optionals in XDR are just unions, and by replacing the optional with an explicit union we can give each of the arms of the union names that clearly describe the behavior we intend in that 'no value' case.

This change is binary compatible.

### Merging

This change is intended to be merged to the `next` branch after May 2nd, such that it falls into the development stream of the next release, not the current release.


Stellar-transaction.x
```diff
@@ -540,9 +540,23 @@ struct AddressWithNonce
...
+case AUTHORIZER_SOURCE_ACCOUNT:
+    void;
+case AUTHORIZER_ADDRESS:
+    AddressWithNonce address;
+};
+
 struct ContractAuth
 {
-    AddressWithNonce* addressWithNonce; // not present for invoker
+    Authorizer authorizer;
```
dmkozh, line 559: Would it make sense to also move `signatureArgs` to `Authorizer`? This will auto-ensure the invariant of having no signatures for the invoker.

leighmcculloch: That is less clear to me.

A reason to keep signatureArgs where it is, is that it is completely reasonable to say that some types of authorizers provide no signatures.

A motivation to move signatureArgs would be if we expect future types of authorizers to be able to pass a different structure as a signature. However, the downstream impact of this would be significant since everything downstream, including custom account contracts, assumes signatures are a vector.

So it sounds good, but not sure if the benefit is real. I think that change can be done and considered separately. This change is purely a naming change and is otherwise binary compatible with the existing XDR.

dmkozh: Ok, I don't have a strong opinion on this, just a thought.

dmkozh (approved):

Stellar-transaction.x
leighmcculloch: I reckon we should do it. At minimum it better communicates the source account case. I'll open a second PR with it.

leighmcculloch: Opened PR with the change at https://github.com/stellar/stellar-xdr/pull/89.

PR #89, opened 2023-04-28
Move ContractAuth signatureArgs into Authorizer

leighmcculloch:
### What
Move ContractAuth signatureArgs into Authorizer.

### Why
Signatures live inside the ContractAuth type. The ContractAuth type defines an authorization for a variety of authorizers. At the moment that includes source accounts and addresses. Source accounts don't have signatures, since the authorization is provided by the fact that the invocation is already signed by the transaction being signed by the source account.

@dmkozh pointed out in https://github.com/stellar/stellar-xdr/pull/88#discussion_r1180781614 that given that no signatures are required when the authorizer is the source account, it might make more sense for the signature to live in the authorizer.

For developers constructing the ContractAuth type, moving the signatures into the AddressAuthorizer will more clearly communicate the intent that the field is only to be used for that authorizer type.

### Merging

This change is intended to be merged to the `next` branch after #88.


dmkozh (approved):

Stellar-transaction.x
```diff
@@ -534,17 +534,31 @@ struct AuthorizedInvocation
     AuthorizedInvocation subInvocations<>;
 };
 
-struct AddressWithNonce
+struct AddressAuthorizer
```
leighmcculloch: Maybe we should reanme `Authorizer` to `Authorization`, and rename this to `AddressAuthorization` given it contains the signatures?

leighmcculloch, line 537: Maybe we should reanme `Authorizer` to `Authorization`, and rename this to `AddressAuthorization` given it contains the signatures?

leighmcculloch, line 537: Maybe we should rename `Authorizer` to `Authorization`, and rename this to `AddressAuthorization` given it contains the signatures?

dmkozh: Yeah, I'd slightly prefer `Authorization` .

dmkozh (approved):

PR #93, opened 2023-05-16
State expiration

SirTyson:
These changes provide full support for state expiration. The includes adding types and expiration data to LedgerEntries, adding lifetime extension fields to the footprint, and additional expiration related network configurations.


Stellar-contract-config-setting.x
```diff
@@ -152,7 +152,11 @@ enum ConfigSettingID
     CONFIG_SETTING_CONTRACT_COST_PARAMS_CPU_INSTRUCTIONS = 6,
     CONFIG_SETTING_CONTRACT_COST_PARAMS_MEMORY_BYTES = 7,
     CONFIG_SETTING_CONTRACT_DATA_KEY_SIZE_BYTES = 8,
-    CONFIG_SETTING_CONTRACT_DATA_ENTRY_SIZE_BYTES = 9
+    CONFIG_SETTING_CONTRACT_DATA_ENTRY_SIZE_BYTES = 9,
+    CONFIG_SETTING_MAXIMUM_ENTRY_LIFETIME = 10,
```
dmkozh: Would it make sense to collapse this into a single struct, like `LedgerEntryLifetimeSettings` or something like that?

Stellar-ledger-entries.x
```diff
@@ -602,11 +641,27 @@ case CONTRACT_DATA:
     {
         Hash contractID;
         SCVal key;
+        ContractDataType type;
```
dmkozh, line 640: This basically makes it possible to have 3 instances of the same key with different lifetimes. Is that expect?

Stellar-ledger-entries.x
```diff
@@ -602,11 +641,27 @@ case CONTRACT_DATA:
...
+
+        // ContractLedgerEntryType t is part of key type, but is inside a switch in the actual entry,
```
dmkozh: I'm not sure I understand that, can't we use just the enum here and then use the xdr discriminant when building the key from the value?

Stellar-transaction.x
```diff
@@ -775,11 +775,18 @@ case PRECOND_V2:
     PreconditionsV2 v2;
 };
 
+// If bumpLedgers is not empty, the given key's lifetime is increased by bumpLedgers
+struct FootprintEntry
+{
+    LedgerKey key;
+    uint32* bumpLedgers;
```
dmkozh: I'm not a huge fan of special values, but here it seems to me like setting `bumpLedgers` to 0 by default would have the right semantics and save 1-4 bytes per footprint entry (not sure what is the optional field overhead).

Stellar-ledger-entries.x
SirTyson: Yeah, we need each entry type to be in a separate key space for security purposes. Because `UNIQUE` entries have a strong, proof based guarantee of uniqueness, we need to protect against the case where a `UNIQUE` entry gets archived and a `RESOTRABLE` or `TEMPORARY` entry with the same key overwrites the `UNIQUE` entry because they can be created without a proof. So in this case, you can have 3 entries with the same key exist at the same time so long as they have different types.

dmkozh: This is interesting. It would mean that the contract will have to specify the entry lifetime for every storage operation (including reads).  I don't think there are security concerns - contract controls its storage and thus shadowing a unique entry would just be a bug.

I'm actually not sure what is more of a footgun: being able to shadow a unique entry or being able to accidentally create multiple entries. The latter actually seems somewhat worse to me. Ideally we'd need to provide lifetime only when creating new entries and disallow creating/modifying the entries with inconsistent lifetimes.

SirTyson: I'm not sure what you mean by lifetime here. Do you mean the entry type, like `UNIQUE`, `TEMPORARY`, etc? Assuming so, the issue is we can't consistently disallow creating and modifying entries with inconsistent types. For example, say a buggy contract calls `set(0x123, UNIQUE)` and `set(0x123, TEMP)`. Sometimes this would be disallowed if the original `UNIQUE` entry was still in the Ledger, but if it had expired, then there would be no way of disallowing the creation of the temp entry. At least in my mind `UNIQUE` entries have very strong security invariants, and the only way to maintain those consistently is via a seperate key space.

This may be a bad interface though having to specify the key and type for every operation. This made more sense originally when I imagined that the SDK would define a seperate storage class for each type, i.e. `TempStorage::set`, `UniqueStorage::set`, etc, but this may be too confusing or convoluted.

Stellar-ledger-entries.x
SirTyson: I think the issue is that for switches, `XDR` puts in some overhead, so the underlying struct for the `LedgerKey` union would be different than that of the `LedgerEntry` if we just had the enum without the shell switch. This may not be a big deal, but for every other type it looked like the `LedgerKey` was exactly the same as the beginning of the corresponding `LedgerEntry`, so I maintained that here but it might not be necessary.

dmkozh: I'm not sure if there is any particular reason for these to be matching - not sure if that's a requirement and not a coincedence.

Stellar-ledger-entries.x
dmkozh: Yeah, I meant entry type.  Security guarantees are really the contract responsibility. It's not immediately obvious if the bugs caused by creating two entries with 'almost' the same key (besides type) are better or worse than bugs caused by re-creating entry with a different type. I guess if we have 3 storage kinds, the UX won't be as confusing, but it still seems pretty clunky. I guess I'd like to see more opinions on this.

graydon (approved):

graydon:
@dmkozh I think @SirTyson addressed your comments here -- can you mark those as resolved so we can merge this?

dmkozh (approved):

dmkozh:
Just FYI since my auth changes are merged now, state expiration would be blocked on https://github.com/stellar/rs-soroban-env/pull/823

Stellar-ledger.x
```diff
@@ -490,11 +490,46 @@ struct LedgerCloseMetaV1
...
+    uint64 totalByteSizeOfBucketList;
+
+    // Expired temp keys that are being evicted at this ledger.
+    LedgerKey evictedTemporaryLedgerKeys<>;
+
+    // Expired restorable ledger entries that are being
+    // evicted at this ledger.
+    LedgerEntry evictedRestorableLedgerEntries<>;
+
+    // We forgot to add an ExtensionPoint in v1 but at least
```
sisuresh: Should the extension point be at the top like [this](https://github.com/stellar/stellar-xdr/blob/07e9e82b8d60244596214b4d327506d791d35ad3/Stellar-ledger-entries.x#L139-L141) for more flexibility (the comment explains why).

graydon: Oh, sure. Moved to the top of the struct.

sisuresh (approved):

PR #100, opened 2023-06-02
Make `returnValue` optional in meta.

dmkozh:
This is meaningless for non-Soroban txs.

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -398,21 +398,31 @@ struct DiagnosticEvent
     ContractEvent event;
 };
 
-struct TransactionMetaV3
+struct SorobanTransactionMeta 
 {
-    LedgerEntryChanges txChangesBefore; // tx level changes before operations
-                                        // are applied if any
-    OperationMeta operations<>;         // meta for each operation
-    LedgerEntryChanges txChangesAfter;  // tx level changes after operations are
-                                        // applied if any
+    ExtensionPoint ext;
+
     ContractEvent events<>;             // custom events populated by the
                                         // contracts themselves.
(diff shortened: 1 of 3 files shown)


Stellar-ledger.x
```diff
@@ -407,7 +407,7 @@ struct TransactionMetaV3
                                         // applied if any
     ContractEvent events<>;             // custom events populated by the
                                         // contracts themselves.
-    SCVal returnValue;                  // return value of the invocation.
+    SCVal* returnValue;                 // return value of the host fn invocation (if any)
```
leighmcculloch: Should the Soroban elements be grouped into a struct that as a whole is `None` for other operations? It feels like a data smell that we have multiple fields that are only relevant in one situation.

dmkozh: Yeah, that probably would be better. What do you think @sisuresh ?

sisuresh: Yeah that'll work well now that we don't hash meta components. Related to this, should we include an `ExtensionPoint` at the top of `TransactionMetaV3` and/or in this new struct?

dmkozh: I don't have a strong preference, but since meta itself is an 'extension', do we really need it? Or do you want this to be the final meta version and all the future extensions to happen via extension point? But then if we want to remove something, things become confusing...

sisuresh: Only reason I mentioned it is because a V4 will require us to copy everything over. If we want to remove a field, we'll have to do a V4 which I guess would get a little confusing.

dmkozh: Ok, added the extension point and updated meta to use soroban-specific struct.

sisuresh (approved):

PR #101, opened 2023-06-07
Add execution lanes config

marta-lokhova:
Per https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-07.md#xdr-changes

### DIFF
--- a/Stellar-contract-config-setting.x
+++ b/Stellar-contract-config-setting.x
@@ -1,6 +1,13 @@
 %#include "xdr/Stellar-types.h"
 
 namespace stellar {
+// General “smart contract execution lane” settings
+struct ConfigSettingContractExecutionLanesV0
+{
+    // maximum number of “smart” transactions per ledger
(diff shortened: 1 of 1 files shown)


Stellar-contract-config-setting.x
```diff
@@ -1,6 +1,13 @@
 %#include "xdr/Stellar-types.h"
 
 namespace stellar {
+// General “smart contract execution lane” settings
+struct ConfigSettingContractExecutionLanesV0
+{
+    // maximum number of “smart” transactions per ledger
```
dmkozh: nit: `Soroban` would be more appropriate (we've been trying to use term `Soroban` more in core).

dmkozh (approved):

jayz22 (approved):

Stellar-contract-config-setting.x
marta-lokhova: sg, fixed

PR #104, opened 2023-06-13
Adds expiration iterator

SirTyson:
This adds an `ExpirationIterator` for managing how core emits expiration meta. We want to limit the amount of entries we expire in a given ledger to not strain downstream systems. This is necessary because there can be an arbitrary number of entries that can have a given expiration ledger, so we have to send meta in batches. The idea of the `ExpirationIterator` is that we gradually scan the entire BucketList. On each ledger close, we scan `maxBytesToScan` megabytes of the BucketList and expire up to `maxEntriesToExpire` entries. These two configuration options ensure that we keep up to date with expired entries while not slowing down close time due to too much disk IO or overburdening downstream systems.

To keep track of where we are in the BucketList while scanning, we write the `ExpirationIterator` on every ledger close. It's not really a `ConfigSettingEntry` but that was the most convenient place to put it. We have to store it somewhere on the BucketList so new nodes joining the network know where to start expiration, and it also needs to be updated on each ledger close. I think `ConfigSettingEntry` is the best way to do that currently even though semantically the iterator is not a config setting.

In previous proposals, we discussed creating a new type of Bucket called the "expiration" Bucket that is produced on merges. We would iterate through specific key bounds on a given expiration bucket based on some modulo arithmetic on the current ledger number. This seemed complicated, and adding a new bucket type was controversial. This approach should be much more straightforward, although we may have to tune the `maxEntriesToExpire` and `maxBytesToScan` values over time to not lag behind too much w.r.t. expiring data.

### DIFF
--- a/Stellar-contract-config-setting.x
+++ b/Stellar-contract-config-setting.x
@@ -47,7 +47,7 @@ struct ConfigSettingContractLedgerCostV0
     int64 feeReadLedgerEntry;  // Fee per ledger entry read
     int64 feeWriteLedgerEntry; // Fee per ledger entry write
 
-    int64 feeRead1KB;  // Fee for reading 1KB    
+    int64 feeRead1KB;  // Fee for reading 1KB
     int64 feeWrite1KB; // Fee for writing 1KB
 
     // Bucket list fees grow slowly up to that size
@@ -138,7 +138,7 @@ enum ContractCostType {
 struct ContractCostParamEntry {
     // use `ext` to add more terms (e.g. higher order polynomials) in the future
     ExtensionPoint ext;
-    
+
     int64 constTerm;
     int64 linearTerm;
 };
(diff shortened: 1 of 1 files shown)


dmkozh (approved):

Stellar-contract-config-setting.x
```diff
@@ -153,6 +153,25 @@ struct StateExpirationSettings {
     int64 restorableRentRateDenominator;
     int64 tempRentRateDenominator;
 
+    // max number of entries that emit expiration meta in a single ledger
+    uint32 maxEntriesToExpire;
+
+    // maximum bytes to scan the BucketList for expired entries in a single ledger, in megabytes
+    uint32 maxBytesToScan;
+
+    union switch (int v)
```
dmkozh: For extensions please use `ExtensionPoint` and put it at the top of the entry.

Also, config extensions are kind of unnecessary, aren't they? You could always just add one more entry type (that's why most of the entries aren't extensible).

jayz22: > For extensions please use ExtensionPoint and put it at the top of the entry.

I've seen all the ExtensionPoints moved to top of the entry and been wondering. Is it just stylistic or are their other reasons (such as maintaining a certain sorting order)? Most of the pre-existing entries have `ext` as the last field.

leighmcculloch: @jayz22 We started placing extension points at the beginning because it defers the choice of:
- extension points being dangling where v2 extends off v1
- or, new versions being an entire replacement since it'll allow us to change the type to be a union in a binary compatible way

jayz22: Hmm, still not sure I understand why putting it at the beginning instead of at the end helps with deferring the choice.

dmkozh: The main reason would be the possibility of converting extension point to union.

dmkozh (approved):

Stellar-contract-config-setting.x
```diff
@@ -158,26 +158,29 @@ enum ContractCostType {
...
+    int64 persistentRentRateDenominator;
     int64 tempRentRateDenominator;
 
-    union switch (int v)
-    {
-    case 0:
-        void;
-    } ext;
+    // max number of entries that emit expiration meta in a single ledger
+    uint32 maxEntriesToExpire;
```
jayz22, line 177: ~What happens if number of expiring entries exceeds this? They still all expire but some of them don't emit meta?~
Never mind, your PR description is pretty clear, sorry about that.

PR #106, opened 2023-06-13
Expiration bump op

sisuresh:


Stellar-transaction.x
```diff
@@ -578,6 +579,22 @@ struct InvokeHostFunctionOp
     SorobanAuthorizationEntry auth<>;
 };
 
+enum BumpExpirationType
```
SirTyson, line 588: What would a "non-uniform" rent bump look like? Does uniform here mean that every key is bumped by the same amount of ledgers?

SirTyson (approved):

Stellar-transaction.x
```diff
@@ -578,6 +579,22 @@ struct InvokeHostFunctionOp
...
+{
+    BUMP_EXPIRATION_UNIFORM = 0
+};
+
+/* Bump the expiration ledger of the entries specified in the readOnly footprint
```
dmkozh, line 593: Is doing a bump via RO entries more efficient than via RW entries? I suppose it is (because we should write the expiration entries only), but just wanted to double-check.

Stellar-transaction.x
sisuresh: It means that every entry will be bumped to expire at that ledger (or later if expiration ledger was longer). A non-uniform one would require you to specify a 1 to 1 mapping of keys to ledgersToExpire. There are three ways to do this, but none are great - 
1. Specify the keys in the op, duplicating the keys in the footprint.
2. Force the footprint to be empty and set the keys in the op.
3. Require a vector of integers in the op that maps by index to the entries in the footprint.

Stellar-transaction.x
```diff
@@ -578,6 +579,22 @@ struct InvokeHostFunctionOp
...
+   so they'll expire at least ledgersToExpire ledgers from lcl.
+    Threshold: low
```
dmkozh: Doesn't this op allow spending potentially high amounts of account funds? I'd say it should probably be medium - it's not that different from a payment...

Stellar-transaction.x
SirTyson: Yeah, RO should always be more efficient

sisuresh: Yeah it would also be inaccurate since we wouldn't be writing those entries (I should note though that the current implementation doesn't use EXTENSION_ENTRIES, but that change will be made soon).

Stellar-transaction.x
sisuresh: I made this low because InvokeHostFunctionOp is also low. Should they both be medium instead?

dmkozh: Hmm, that's a bug then. Invoke host fn should definitely be medium.

sisuresh: The [CAP](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0046-04.md) specifically mentions that InvokeHostFunctionOp should be low, but I don't remember Jon's reason for it (I do remember discussing this with him though).

dmkozh: That doesn't really make any sense now (maybe it did in pre-SAC world). Payment is medium threshold and we don't want to relax that with SAC.

sisuresh: Hmm what you're saying makes sense. I'm think this made sense because account auth in the contract used to default to medium threshold back then. I'll make the change here and in core.

dmkozh (approved):

PR #107, opened 2023-06-14
Add contract metrics to the transaction meta

jayz22:
xdr change for https://github.com/stellar/stellar-core/issues/3759
Initially only include the runtime resource consumption: cpu and memory. 
Gradually we can add more.

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -398,6 +398,13 @@ struct DiagnosticEvent
     ContractEvent event;
 };
 
+struct ContractMetrics
+{
+    ExtensionPoint ext;
+    int64 cpuInsnsConsumed;
+    int64 memBytesConsumed;
+};
+
 struct SorobanTransactionMeta 
 {
     ExtensionPoint ext;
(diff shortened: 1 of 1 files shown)


Stellar-ledger.x
```diff
@@ -398,6 +398,13 @@ struct DiagnosticEvent
     ContractEvent event;
 };
 
+struct ContractMetrics
+{
+    ExtensionPoint ext;
+    int64 cpuInsnsConsumed;
```
sisuresh, line 404: Should these be unsigned?

jayz22: Maybe. I don't think it matters since many of the [configs](https://github.com/stellar/stellar-xdr/blob/next/Stellar-contract-config-setting.x) are also i64 for max generality. I don't feel too strongly either way.

sisuresh: The actual tracking values in core are unsigned, so I think it's safer to make these unsigned as well.

jayz22:
No longer needed, we will do them as diagnostic events as per issue.

PR #119, opened 2023-06-22
Make `SCContractInstance` a pointer in `ScVal`.

dmkozh:
This makes `SCVal` smaller at cost of making instance entries slightly larger, which should be a net win.

This also makes it easier to use the storage in `ScContractInstance` as it can't be in two 'empty' states (null or empty map).

### DIFF
--- a/Stellar-contract.x
+++ b/Stellar-contract.x
@@ -186,7 +186,7 @@ struct SCNonceKey {
 
 struct SCContractInstance {
     ContractExecutable executable;
-    SCMap* storage;
+    SCMap storage;
 };
 
 union SCVal switch (SCValType type)
@@ -248,7 +248,7 @@ case SCV_LEDGER_KEY_NONCE:
     SCNonceKey nonce_key;
 
 case SCV_CONTRACT_INSTANCE:
-    SCContractInstance instance;
+    SCContractInstance* instance;
 };
 
 struct SCMapEntry


Stellar-contract.x
```diff
@@ -248,7 +248,7 @@ case SCV_LEDGER_KEY_NONCE:
     SCNonceKey nonce_key;
 
 case SCV_CONTRACT_INSTANCE:
-    SCContractInstance instance;
+    SCContractInstance* instance;
```
sisuresh, line 251: This is a little weird because a null instance is never valid right?

dmkozh: Sure, but that's fine. Map and vec are implemented in the same way and this one is not even user-visible.

jayz22:
> This makes SCVal smaller at cost of making instance entries slightly larger

I'm struggling to understand why that is. Would you mind elaborate it a little bit? Why does making `SCContractInstance` optional under ScVal reduce its size? Optional data in Rust just enum which is the same size as the discriminant?
Also not understanding how does it makes the instance entry larger. It's just reshuffling which level the enum happens.

dmkozh:
> > This makes SCVal smaller at cost of making instance entries slightly larger
> 
> I'm struggling to understand why that is. Would you mind elaborate it a little bit? Why does making `SCContractInstance` optional under ScVal reduce its size? Optional data in Rust just enum which is the same size as the discriminant? Also not understanding how does it makes the instance entry larger. It's just reshuffling which level the enum happens.

You are right, this apparently doesn't help, which is really annoying. I would expect option to be treated as 0 to 1 element vec (so the shallow copy would be cheap), not as the whole underlying value.

I would still consider this change though, as the second reason stays valid (it is more convenient to have the whole instance wrapped in option instead of handling two storage states separately).

dmkozh:
> > > This makes SCVal smaller at cost of making instance entries slightly larger
> > 
> > 
> > I'm struggling to understand why that is. Would you mind elaborate it a little bit? Why does making `SCContractInstance` optional under ScVal reduce its size? Optional data in Rust just enum which is the same size as the discriminant? Also not understanding how does it makes the instance entry larger. It's just reshuffling which level the enum happens.
> 
> You are right, this apparently doesn't help, which is really annoying. I would expect option to be treated as 0 to 1 element vec (so the shallow copy would be cheap), not as the whole underlying value.
> 
> I would still consider this change though, as the second reason stays valid (it is more convenient to have the whole instance wrapped in option instead of handling two storage states separately).

IIUC we can't have no pointer here for the same reason we had to have pointers for scval/scmap (xdrpp recursion support).

dmkozh:
Yeah, I've just checked using a vector for this and `ScVal` size comes back to 40 bytes. Maybe we just need to be smarter about metered clone implementation for scval. Or maybe it's not significant enough to worry about...

jayz22:
> Yeah, I've just checked using a vector for this and `ScVal` size comes back to 40 bytes. Maybe we just need to be smarter about metered clone implementation for scval. Or maybe it's not significant enough to worry about...

Sure if you are using a small member type in the structure, the structure size reduces. 
However, the actual size (e.g. being charged for metered_clone) won't change, it is just pushed into the heap and still will be charged. Plus the extra 24 bytes of the ScVec that wraps it. So it will end up being more. 

I don't think this is worth the effort from the data size perspective. But I get it from the other reason ("no empty") perspective.

dmkozh:
> Sure if you are using a small member type in the structure, the structure size reduces.
However, the actual size (e.g. being charged for metered_clone) won't change, it is just pushed into the heap and still will be charged. Plus the extra 24 bytes of the ScVec that wraps it. So it will end up being more

I mean a regular XDR vec, not ScVec. This should reduce the shallow clone copy for every ScVal, as I intended with this PR. That's ugly though, so I'm not sure if that's worth it.

jayz22:
> I mean a regular XDR vec, not ScVec. This should reduce the shallow clone copy for every ScVal

I think it's  be same (vec or ScVec), you are still just pushing the memory of the structure into the heap. For cloning the ScVal it will clone both the shallow ScVal and the heap memory so in the end it will cost a few bytes more.

dmkozh:
> I think it's be same (vec or ScVec), you are still just pushing the memory of the structure into the heap. For cloning the ScVal it will clone both the shallow ScVal and the heap memory so in the end it will cost a few bytes more.

The issue is declared size of the `ScVal`. Without `SCContractInstance` it was just 40. With it it's 64. Using XDR vec brings the size back to 40. This means that for *every* clone of *every* `ScVal` we charge for 64 bytes copy. This might be little in absolute terms, but it's still +50% in relative terms, which is a bit concerning (depends on how much we clone these really). This has nothing to do with the actual allocations needed for contract instance.

jayz22:
> This means that for every clone of every ScVal we charge for 64 bytes copy

Every clone of ScVal also clones its sub-structure right?

dmkozh:
> Every clone of ScVal also clones its sub-structure right?

Sure, but substructure depends on the active variant, right? Ironically, we almost never need to clone contract instances, which makes this even more annoying.

jayz22:
> > Every clone of ScVal also clones its sub-structure right?
> 
> Sure, but substructure depends on the active variant, right? Ironically, we almost never need to clone contract instances, which makes this even more annoying.

I thought you were concerned with the actual cost of cloning, not metered cost. 

Overcharging (metering) a few bytes more for the empty variant should not be a concern. The entire metering framework assumes we will have to over-charge for things to account for the worst case, as long as the overcharging is known and calculated (e.g. non-recursive). I think we should be fine with overcharging a few extra pointer size worth of bytes per structure. 

Also want to point out, cloning ScVal is not going to happen too much anyway. Since all the host entities are efficiently stored as host objects, we will only be cloning the raw xdr objects when crossing host boundaries (e.g cloning events out of the host) which happens for known and calculable times.  

The actual storage-wise, storing the empty variant as a `Vec` costs 24 bytes more overhead than an `Option::None`, which is the real cost and I think is what we should be optimizing for.

Stellar-contract.x
leighmcculloch: Is this union / optional required because of a cyclic reference or is the value in fact optional?

leighmcculloch: Ah I see this is the same as why Map and Vec are optional.

dmkozh:
> Overcharging (metering) a few bytes more for the empty variant should not be a concern. 

This is actually exactly what I'm concerned about - of course, the metering requirement should be that we don't exceed realtime performance much, but we also want to be fair to the users. Since we don't clone ScVals much I agree that we don't need to try optimizing for this particular issue, but in general I think we should try to be precise with metering and avoid overcharging if possible/feasible.

jayz22:
> > Overcharging (metering) a few bytes more for the empty variant should not be a concern.
> 
> This is actually exactly what I'm concerned about - of course, the metering requirement should be that we don't exceed realtime performance much, but we also want to be fair to the users. Since we don't clone ScVals much I agree that we don't need to try optimizing for this particular issue, but in general I think we should try to be precise with metering and avoid overcharging if possible/feasible.

Yep, we will hopefully have a better assessment once we have the end-to-end testing, metrics and real world stats monitoring setup for the testnet.

PR #120, opened 2023-06-22
XDR for RestoreFootprintOp

sisuresh:


Stellar-transaction.x
```diff
@@ -602,6 +603,22 @@ case BUMP_FOOTPRINT_EXPIRATION_UNIFORM:
     uint32 ledgersToExpire;
 };
 
+enum RestoreFootprintType
+{
+    RESTORE_FOOTPRINT_V1 = 0
```
sisuresh: This is a little weird, but I think we should make the op a union in case we need to update how restoration works for evicted entries.

SirTyson (approved):
LGTM

Stellar-transaction.x
dmkozh: Maybe just leave an extension point for now? It will achieve the same without much verbosity.

sisuresh: I was thinking about that but wasn't sure if it'd work. If we use an extension point and then switch to a union later, can the type in the union be an enum?

sisuresh: Actually I think you're right. It would still be binary compatible.

dmkozh: IIUC yes (and this is why we define extension points at the top of the structs - that provides binary compatibility with unions.

sisuresh: I updated both the bump and restore ops to use extension points.

Stellar-transaction.x
```diff
@@ -585,23 +586,28 @@ struct InvokeHostFunctionOp
...
    so they'll expire at least ledgersToExpire ledgers from lcl.
 
     Threshold: med
     Result: BumpFootprintExpirationResult
 */
-union BumpFootprintExpirationOp switch (BumpFootprintExpirationType type)
+struct BumpFootprintExpirationOp
 {
-case BUMP_FOOTPRINT_EXPIRATION_UNIFORM:
+    ExtensionPoint
```
dmkozh: This looks broken to me

dmkozh (approved):

PR #124, opened 2023-07-17
Soroban tx set size upgrade support and XDR cleanup.

dmkozh:
- Use dedicated struct for `InvokeContract` host fn
- Replace `ScVec` with `SCVal` or `SCVal<>` depending on the context

### DIFF
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -503,10 +503,16 @@ struct CreateContractArgs
     ContractExecutable executable;
 };
 
+struct InvokeContractArgs {
+    SCAddress contractAddress;
+    SCSymbol functionName;
+    SCVal args<>;
+};
+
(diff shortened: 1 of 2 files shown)


sisuresh (approved):

Stellar-transaction.x
```diff
@@ -503,10 +503,16 @@ struct CreateContractArgs
     ContractExecutable executable;
 };
 
+struct InvokeContractArgs {
+    SCAddress contractAddress;
```
2opremio, line 507: Shouldn't this just be a hash? In which cases can this be an account address?

dmkozh: This is for consistency and extensibility. All the contract interactions use address now, so it would be weird to omit it in one place. This also allows us to add functionality to classic accounts if we ever want that (for example, imagine some subscription interface).

PR #126, opened 2023-07-20
Remove incorrect comment about Soroban u256 representation

brson:
u256 is no longer represented as an array of bytes. It is represented as u64s similarly to the u128 case, so there's no comparison needed here.


tomerweller (approved):

tomerweller:
thanks!

PR #130, opened 2023-07-31
Adds eviction iterator

SirTyson:
Adds iterator to keep track of place in eviction scans. On each ledger close, a small portion of a given bucket is scanned, and any expired entries are evicted. This iterator keeps track of the current scan location.


dmkozh (commented):
Just FYI you'll need to wait for a bit before you'll be able to use this in core (there are some XDR changes that still need to be adopted).

dmkozh (approved):

PR #132, opened 2023-08-04
Add max bump error

sisuresh:
For the core side of https://github.com/stellar/rs-soroban-env/issues/973.


sisuresh:
Nvm this isn't necessary. The bump is already capped at the max during validation.

PR #137, opened 2023-08-16
Expiration Entry rework

SirTyson:
WIP Expiration Entry rework.

### DIFF
--- a/Stellar-ledger-entries.x
+++ b/Stellar-ledger-entries.x
@@ -100,7 +100,8 @@ enum LedgerEntryType
     LIQUIDITY_POOL = 5,
     CONTRACT_DATA = 6,
     CONTRACT_CODE = 7,
-    CONFIG_SETTING = 8
+    CONFIG_SETTING = 8,
+    EXPIRATION = 9
 };
 
 struct Signer
@@ -493,58 +494,32 @@ struct LiquidityPoolEntry
     body;
 };
 
-enum ContractEntryBodyType {
-    DATA_ENTRY = 0,
-    EXPIRATION_EXTENSION = 1
-};
-
-const MASK_CONTRACT_DATA_FLAGS_V20 = 0x1;
-
-enum ContractDataFlags {
-    // When set, the given entry does not recieve automatic expiration bumps
-    // on access. Note that entries can still be bumped manually via the footprint.
-    NO_AUTOBUMP = 0x1
-};
-
 enum ContractDataDurability {
     TEMPORARY = 0,
     PERSISTENT = 1
 };
 
 struct ContractDataEntry {
+    ExtensionPoint ext;
+
     SCAddress contract;
     SCVal key;
     ContractDataDurability durability;
-
-    union switch (ContractEntryBodyType bodyType)
-    {
-    case DATA_ENTRY:
-    struct
-    {
-        uint32 flags;
-        SCVal val;
-    } data;
-    case EXPIRATION_EXTENSION:
-        void;
-    } body;
-
-    uint32 expirationLedgerSeq;
+    SCVal val;
 };
 
 struct ContractCodeEntry {
     ExtensionPoint ext;
 
     Hash hash;
-    union switch (ContractEntryBodyType bodyType)
-    {
-    case DATA_ENTRY:
-        opaque code<>;
-    case EXPIRATION_EXTENSION:
-        void;
-    } body;
-
-    uint32 expirationLedgerSeq;
+    opaque code<>;
 };
+
+struct ExpirationEntry {
+    // Hash of the LedgerKey that is associated with this ExpirationEntry
+    Hash keyHash;
+    uint32 expirationLedgerSeq;
+}
 
 
 struct LedgerEntryExtensionV1
@@ -583,6 +558,8 @@ struct LedgerEntry
         ContractCodeEntry contractCode;
     case CONFIG_SETTING:
         ConfigSettingEntry configSetting;
+    case EXPIRATION:
+        ExpirationEntry expiration;
     }
     data;
 
@@ -643,19 +620,23 @@ case CONTRACT_DATA:
         SCAddress contract;
         SCVal key;
         ContractDataDurability durability;
-        ContractEntryBodyType bodyType;
     } contractData;
 case CONTRACT_CODE:
     struct
     {
         Hash hash;
-        ContractEntryBodyType bodyType;
     } contractCode;
 case CONFIG_SETTING:
     struct
     {
         ConfigSettingID configSettingID;
     } configSetting;
+case EXPIRATION:
+    struct
+    {
+        // Hash of the LedgerKey that is associated with this ExpirationEntry
+        Hash keyHash;
+    } expiration;
 };
 
 // list of all envelope types used in the application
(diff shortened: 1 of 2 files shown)


Stellar-ledger-entries.x
```diff
@@ -643,19 +621,22 @@ case CONTRACT_DATA:
...
     } contractCode;
 case CONFIG_SETTING:
     struct
     {
         ConfigSettingID configSettingID;
     } configSetting;
+case EXPIRATION:
+    struct
+    {
+        Hash keyHash;
```
Shaptic, line 638: do you need a corresponding preimage variant, e.g. `HashIdPreimageLedgerKey`? not sure if you guys always have a 1:1 relationship between a hash property and a corresponding preimage structure

2opremio:
Splitting the data and expiration in separate ledger entries probably calls for reconsidering the XDR of the bump/restore footprint operations.

The ledger entries to bump/restore were implicitly supplied through keys in the footprint from the soroban transaction data. But now I guess that the final (i.e. preflighted) footprint will contain two entry keys (one for the data and one for the expiration), making it (even more) odd to use it as an implicit parameter.

I would suggest to add the ContractData/ContractCode keys explicitly to the operation bodies. As a side effect this will allow us to restrict the keys passed to ContractData/ContractCode through typing (e.g. defining a new union, ExpirableLedgerEntryKey, which only allows passing those keys).

2opremio:
In fact, we may even want to reconsider the bump/restore operation names. Maybe `BumpContractLedgerEntries` and `RestoreContractLedgerEntries` would make more sense if we decided not to use the footprint as input anymore.

Stellar-ledger-entries.x
dmkozh: It can be just hash of `LedgerKey`. Preimages are needed for external stuff, but for internal stuff I don't think we need to distinguish hash namespaces.

Shaptic: Fine by me - a comment here would be appreciated, then :+1:

dmkozh:
> I would suggest to add the ContractData/ContractCode keys explicitly to the operation bodies.

This PR in no way changes the reason for using the footprint for these ops, namely the need for consistency between how the footprints are treated for different operations. But whether expiration entries are present in the footprint is still up to debate.

SirTyson:
> > I would suggest to add the ContractData/ContractCode keys explicitly to the operation bodies.
> 
> This PR in no way changes the reason for using the footprint for these ops, namely the need for consistency between how the footprints are treated for different operations. But whether expiration entries are present in the footprint is still up to debate.

Now that we've hashed out a concurrency model, ContractData/ContractCode keys need to be in the RO set for bumps and restores, the associated ExpirationEntry key should not be present in the footprint.

dmkozh:
> Now that we've hashed out a concurrency model, ContractData/ContractCode keys need to be in the RO set for bumps

Just to add a bit more on that, they *have* to be in the footprint in order to guarantee the correct data dependency semantics (I don't think we have the time/need to move bump ops to a separate stage).

Stellar-ledger-entries.x
```diff
@@ -493,58 +494,32 @@ struct LiquidityPoolEntry
...
-    case DATA_ENTRY:
-        opaque code<>;
-    case EXPIRATION_EXTENSION:
-        void;
-    } body;
+    opaque code<>;
+};
 
+struct ExpirationEntry {
+    Hash keyHash;
```
2opremio, line 520: Not having a way to get the Contract Data/Code ledger entry from the ExpirationEntry makes things like https://github.com/stellar/soroban-tools/pull/865 quite hard.

The storage won't be able to intercept queries to ExpirationEntries and track what matching Contract Data/Code entry requires restoring.

i.e. I wouldn't know how to implement this https://github.com/stellar/soroban-tools/pull/865/files#diff-f9ab882008e7415ec69e8078b2d91d0a16b9b584cef6c5fb04e55bccfa26e68dR55-R93

CC @dmkozh

2opremio: Thus, I would suggest storing the full ledger key instead of the hash

dmkozh: Contracts don't access expiration entries directly. Since we have expiration in a separate entry now, we might need to revisit the snapshot interface, maybe by adding `expiration(LedgerKey)`. But this is still easy to implement by computing the hash of the key before the access.

2opremio: Contracts currently don't, but does that mean that they (or the host) won't ever access them?

SirTyson: Yes, the way we've designed our concurrency model, it's a strong requirement that `ExpirationEntry` will never be directly accessed, exposed to contract developers, or be in a transaction footprint. Users will always query the lifetime of an entry via `ContractCode/ContractData` key, never directly with `ExpriationEntry` key.

2opremio: ok, then I guess we should also be hidding expiration entries in the `getLedgerEntry` method of soroban-rpc

tsachiherman: One of the thoughts about the getledgerentries was to allow a downstream client to run transaction simulation using a remote ledger state. I don't see us implementing that right away, but don't mind leaving that as an extension to someone else ( without making any commitment ). Do you see any downside for leaving this enabled ?

SirTyson: No downside, I think we should leave this enabled. `getLedgerEntry` is unrelated to the concurrency model for TX application, so it's fine to access these entries there.

tsachiherman: I disagree with Garand about the semantics; the developer ( js ) needs to be aware of the expiration. What Garand meant ( I think.. ) was that we should avoid exposing the key semantic to the developer, and provide the developer with nice APIs that would wrap up that information.

Stellar-ledger-entries.x
```diff
@@ -493,58 +494,32 @@ struct LiquidityPoolEntry
...
-        opaque code<>;
-    case EXPIRATION_EXTENSION:
-        void;
-    } body;
+    opaque code<>;
+};
 
+struct ExpirationEntry {
+    // Hash of the LedgerKey that is associated with this ExpirationEntry
+    Hash keyHash;
```
Shaptic, line 520: wait, isn't the entry supposed to hold the expiration sequence number, while the key holds the hash?

SirTyson: A `LedgerEntry` must contain its `LedgerKey`. We don't store keys and values separately in the BucketList, we only store `LedgerEntry` and cast the `LedgerEntry` to a `LedgerKey` when searching by key.

Shaptic: Oh okay. I think I was confused because I missed subtle little L521 due to the GH comments (right below this thread), so I was confused where the `expirationLedgerSeq` was stored now

Stellar-ledger-entries.x
dmkozh, line 1: Could you please also remove `autobumpLedger` setting?

Stellar-ledger-entries.x
2opremio: @tsachiherman we already disable expired entries in `getLedgerEntries`. Thus I don’t think it makes sense to allow querying the expiration.

But this is a discussion for a different place.

dmkozh (approved):

PR #141, opened 2023-08-25
Add more granular tx and op errors for soroban

sisuresh:


Stellar-transaction.x
```diff
@@ -1790,7 +1790,9 @@ enum InvokeHostFunctionResultCode
     INVOKE_HOST_FUNCTION_TRAPPED = -2,
     INVOKE_HOST_FUNCTION_RESOURCE_LIMIT_EXCEEDED = -3,
     INVOKE_HOST_FUNCTION_ENTRY_EXPIRED = -4,
-    INVOKE_HOST_FUNCTION_INSUFFICIENT_REFUNDABLE_FEE = -5
+    INVOKE_HOST_FUNCTION_INSUFFICIENT_REFUNDABLE_FEE = -5,
+    INVOKE_HOST_FUNCTION_CODE_SIZE_ABOVE_NETWORK_LIMIT = -6,
+    INVOKE_HOST_FUNCTION_DATA_SIZE_ABOVE_NETWORK_LIMIT = -7
```
dmkozh: On one hand, this is a bit risky because the apply-time errors are canonical and thus we won't be able to e.g. change order for checking the entry sizes and total read bytes or even change the fetching order between code and data entries. On the other hand, we could expand `INVOKE_HOST_FUNCTION_RESOURCE_LIMIT_EXCEEDED` as well if we're going for more granular error codes anyway. I don't have a strong opinion on this really, but introducing granular codes for just a couple of cases seems a bit weird.

dmkozh: If we're going for the network limits only here, the events may exceed it as well.

sisuresh: This was just a fix for the inaccurate code, but yeah this isn't consistent with the changes in TransactionFrame. I agree that adding more error codes complicates the operation (I think this was Jon's original point in not having granular error codes in Soroban). Maybe I should just use a single `INVOKE_HOST_FUNCTION_ENTRY_SIZE_ABOVE_NETWORK_LIMIT` error? Or just undo this change? Maybe we just need to document every failure case? 

Btw in addition to the events, memory is also checked against network config.

dmkozh: I think the safest way would be to revert the operation change or maybe distinguish between exceeding the network limit and transaction-level limit. In any case, it would be nice to synthesize a diagnostic event that specifies what has been exceeded and how (good thing is that we don't have to do that immediately). Documentation unfortunately doesn't help debugging much (like one would know that one of 10 different things failed) and granular codes are risky. I'm also curious what @graydon and @MonsieurNicolas think.

Stellar-transaction.x
```diff
@@ -2034,7 +2052,14 @@ struct TransactionResult
...
     case txMALFORMED:
-    case txSOROBAN_RESOURCE_LIMIT_EXCEEDED:
+    case txSOROBAN_INSTRUCTIONS_ABOVE_NETWORK_LIMIT:
+    case txSOROBAN_READ_BYTES_ABOVE_NETWORK_LIMIT:
+    case txSOROBAN_WRITE_BYTES_ABOVE_NETWORK_LIMIT:
+    case txSOROBAN_READ_ENTRIES_ABOVE_NETWORK_LIMIT:
+    case txSOROBAN_WRITE_ENTRIES_ABOVE_NETWORK_LIMIT:
+    case txSOROBAN_TX_SIZE_ABOVE_NETWORK_LIMIT:
+    case txSOROBAN_INVALID_READ_FOOTPRINT:
+    case txSOROBAN_INVALID_WRITE_FOOTPRINT:
```
Shaptic, line 2058: Do we want to, possibly, make these a `SorobanTransactionResult` to encompass all Soroban errors rather than having the `txSOROBAN_` prefix?

Stellar-transaction.x
sisuresh: Yeah I'll just undo the op change and we can add diagnostic events later.

Stellar-transaction.x
sisuresh: I think the error to the end user should still mention soroban (for example `TX_SIZE_ABOVE_NETWORK_LIMIT` could confuse a downstream user into thinking the limit they're looking for applies to classic ops as well). @MonsieurNicolas do you have an opinion on this?

MonsieurNicolas: I think the point that Dima made above (wrt operations) probably applies for tx level errors:
adding error codes like this does not really give enough information for people (end users, not client devs) to do much and forces clients to handle those errors (even though the only thing they can tell end users is "re-run preflight").

For example, when getting `txSOROBAN_INVALID_READ_FOOTPRINT`, people (developers) want to know "which footprint" during their debugging session.

The only error code we need at this layer is probably just a `txSOROBAN_TRY_AGAIN` or something like that (that tells people to re-run preflight).

In https://github.com/stellar/stellar-core/issues/3816#issuecomment-1626371196 what I was suggesting for "transaction submission errors" was to add a new node in the http `tx` json endpoint:
right now `error` contains the `TransactionResult`, we can relatively easily add `diagnostics` (that contains soroban diagnostic events for example). Of course this would only be populated if core has diagnostic events enabled (which is the case for "Soroban-RPC").

This approach is the most flexible and the least disruptive for v1.
In the future, we could decide to "first class" certain error codes (and name them clearly as "tx submission time"), if there are certain scenarios where we expect clients to do something other than "retry".

sisuresh: Most of these errors cannot be solved by re-running preflight, so rpc would need to pass these errors back to the user somehow (cc @2opremio). You are describing an additional "logging" mechanism in `TransactionFrame` where the infomation is passed back in the `tx` endpoint. This will take an additional day, so that should be kept in mind wrt the timeline (@tsachiherman @ire-and-curses).

MonsieurNicolas: Reason I say "re-run preflight" is because those errors should be caught well before submitting to the network and the solution (from a end user point of view), is to retry:
* over limits type of error -> tx should fail during preflight, don't even need to sign it
* bad footprints -> most likely stale footprint, preflight will resolve this

with information from preflight, clients can actually surface a lot more useful information to the end-users.

On the other hand, the later errors get surfaced to end-users, the more frustrating the experience, especially if people don't submit the transaction right away.

sisuresh: > over limits type of error -> tx should fail during preflight, don't even need to sign it

This would require RPC to do the same limit checks as core. Are we planning on adding these @tsachiherman?

tsachiherman: We can definitely do these.. but I would need some help to come up with the optimal "cookbook".
I believe that some of these errors need to be further thought in order for us to design the proper transaction submission flow. The soroban-rpc role here, could be to help the developer at certain points, but we'll need to evaluate the full extent of the potential errors in order to deliver a complete workable solution.
( i.e. the js-sdk would also be part of that.. as well as the dapp itself ).

MonsieurNicolas: My recommendation on this one would be to optimize for v1 for the developer experience:
* lean harder on diagnostic events, potentially exposing them "as-is" like those horrible Java errors to end users (but at least this pushes information from end-users to wallet devs etc and in some cases the errors should a bit cryptic but actionable even for end users).
* this creates a development flow that allows devs to fix their code (including devs in SDKs, platform and core) over time
* iterate over a few releases (not protocol releases) as to close the gap so that we do not ask end users to sign invalid transactions in the first place.

> This would require RPC to do the same limit checks as core. Are we planning on adding these @tsachiherman?

It may make sense to move some of the logic to a shared Rust module to avoid duplication (like what we did for fees from what I understand).

sisuresh: It sounds like these are the steps to take - 
1. Change `txSOROBAN_RESOURCE_LIMIT_EXCEEDED` to something like `txSOROBAN_INVALID`.
2. Add "diagnostic events" for Soroban errors. These aren't like contract diagnostic events that show up in meta. Instead, they're just conveyed in the `tx` endpoint. 
3. Preflight should handle these "fast fail" cases. Ideally we'd use a shared library. The checks are currently in C++ though so we'll need to figure some things out.

graydon: I'll take this for now. @MonsieurNicolas just to be clear do you want to emit the actual XDR of a diagnosticevent (or a JSON-ified view of it) from the `tx` endpoint? Or just something morally equivalent (eg. "a string that's a more detailed message explaining what went wrong")?

MonsieurNicolas: probably what we do for `TransactionResult`: we embed base64 encoded, xdr encoded in a single json field so that we don't have to invent new schemas

2opremio: > For example, when getting `txSOROBAN_INVALID_READ_FOOTPRINT`, people (developers) want to know "which footprint" during their debugging session.
> 
> The only error code we need at this layer is probably just a `txSOROBAN_TRY_AGAIN` or something like that (that tells people to re-run preflight).

Please please please let's not do this. This degrades error codes even further instead of improving them (the purpose of this PR).

Client developers are also users, let's not forget that. And, in case of bugs, errors are helpful to diagnose the situation. We shouldn't assume preflight is perfect nor perfectly matching what core does because it already isn't (e.g. recording vs enforcing mode).

If there is a bug or a mismatch, re-running preflight won't necessarily help and `txSOROBAN_TRY_AGAIN` would be misleading.

I am aware you are also proposing to add diagnostic events to the response, which is great and it would help further but it shouldn't justify degrading error codes even more.

2opremio: Also, if you end up deciding to use a single error code please don't release it until the diagnostic events returned are refined and really detailed.

Otherwise we are going to be in a much worse place than we are now. Even if that situation is fixed eventually it's going to be very painful until it is resolved.

In fact, if you decide to go with a single error code, I would actually propose a transitioning period in which we maintain/expand the error codes and only rely on diagnostic events when we confirm they are good enough.

ire-and-curses: @2opremio and I discussed this a bit more. Here's where we're at:

There are two distinct problems that we're thinking about:

1. client devs need decent info to debug
2. end users need an experience that makes sense

and we have some preferences:

- _platform:_ as client devs we'd rather have more info than less
- _platform:_ we'd rather see any new system put through its paces before leaning on it heavily
- _core:_ we'd rather not add more error codes because it increases the surface area of the protocol
- _core:_ we'd rather not add more error codes because they don't help end users understand what to do in some cases
- _core:_ we'd rather not add more error codes because in a theoretical different workflow with preflight guards most of these errors should not occur
- _core:_ we'd rather add diagnostics rather than error codes because they are lighter weight (don't need protocol upgrades) so can iterate on much faster

So I understand the constraints - they make sense. What I want to ensure is that when we do this:
> Add "diagnostic events" for Soroban errors. These aren't like contract diagnostic events that show up in meta. Instead, they're just conveyed in the tx endpoint.

the outcome is good enough that we can debug effectively on the platform side. This is _really really_ important to keep us on track for the timeline. Let's be clear on the cost of poor information: Each of the present incomprehensible error scenarios is costing both platform and core teams hours of debug time, and adding lag due to time zones.

So here's the ideal problem statement:

If there is a problem with resources I would like to know:

- **What** resource is affected: footprint, read bytes, write bytes etc ...
- **Why:** Is it because the resource is too low? If so, what is the minimum required when detecting the problem? Is it because the resource is too high? What's the maximum?
- **Some detail:** If it's a footprint, what key is affected? Is it from the readonly footprint or from the readwrite footprint? Why is it affected?

Do we think that's achievable with the proposed solution?

sisuresh: > If there is a problem with resources I would like to know:
What resource is affected: footprint, read bytes, write bytes etc ...
Why: Is it because the resource is too low? If so, what is the minimum required when detecting the problem? Is it because the resource is too high? What's the maximum?
Some detail: If it's a footprint, what key is affected? Is it from the readonly footprint or from the readwrite footprint? Why is it affected?
Do we think that's achievable with the proposed solution?

Yeah the diagnostic solution @graydon is adding will allow us to return any information we want, allowing for granular diagnostics.

And to @2opremio's point about degrading error codes, the [plan](https://github.com/stellar/stellar-xdr/pull/143/files) is to just rename the error to make it more accurate (`txSOROBAN_RESOURCE_LIMIT_EXCEEDED` -> `txSOROBAN_INVALID`).

ire-and-curses: That's awesome! If we can vet that we return info that can answer the Qs I posed, this will be huge. Looking forward to seeing this. :)

tsachiherman: I'm not sure that I'm a big fan of hiding all errors behind a txSOROBAN_INVALID and letting the end user read the error string. Having the error string is great - no issues here. However, in case a downstream system would need to make a logical decision - it should be based on error codes rather then error strings. ( i.e. we shouldn't require downstream systems from parsing errors strings. Unlike error codes, error strings can change over time ).

Shaptic: Diagnostics have to be enabled in this case, right? Maybe it'd be nice if there was at least _some_ information that didn't require an opt-in RPC host.

sisuresh: @tsachiherman The errors will use the `DiagnosticEvents` xdr struct (here is an example error from @graydon's core PR). 

George that's an interesting question. In this case, it might make sense to always have them enabled? We shouldn't have that many of these new diagnostic events per tx. Is there a downside to always returning these new events @graydon?

Shaptic: If they are `system` rather than `diagnostic`, then :+1:

2opremio: Diagnostic events are opt out in soroban rpc

graydon: Diagnostic events are _structured_: they don't _just_ have a string / data field. They also have a more-structured topic vector intended for machine filtering, and for errors we are doing the same thing here (at tx validation time) as we do with error diagnostic events emitted at execution time into the meta: the topics array is always `[SCSymbol("error"), SCError(SCErrorType,SCErrorCode)]`.

For example in the PR I just pushed, if you submit a tx that wants more instructions than the network limit, you'll get a diagnostic event during `tx?` endpoint submission that decodes to the following:

```
{ inSuccessfulContractCall = FALSE,
  event = {
    ext = {
      v = 0
    },
    contractID = NULL,
    type = DIAGNOSTIC,
    body = {
      v = 0,
      v0 = {
        topics = [
          { type = SCV_SYMBOL,
            sym = "error" },
          { type = SCV_ERROR,
            error = {
              type = SCE_BUDGET,
              code = SCEC_EXCEEDED_LIMIT
            } }
        ],
        data = {
          type = SCV_VEC,
          vec = [
            { type = SCV_STRING,
              str = "transaction instructions resources exceed network config limit" },
            { type = SCV_U64,
              u64 = 2000000000 },
            { type = SCV_U64,
              u64 = 100000000 }
          ]
        }
      }
    }
  } }
```

> Unlike error codes, error strings can change over time

Just to be clear, this is not an accidental or negative side effect. This is something pursued intentionally here, by choice. The ability to change these over time is one of the reasons _why_ we want to use something like diagnostic events -- which are not hashed into the canonical history of the ledger -- rather than error codes. Because error codes impose a higher level of rigidity and maintenance burden, prevent evolution of the implementation or protocol. Any error code we emit gets hashed into the history and we must reproduce it bit-identically in every past circumstance that ever produced it. We want to limit the quantity of things like that.

Some other reasons to prefer `DiagnosticEvent`s that might appeal to people outside the core team:
  - They're more open-ended, there's really no maintenance burden on making them as specific as we can, every single line of code that fails one way or another can have its own diagnostic event message, it's not like they chew up scarce numbers or produce annoying error lists in XDR that spam all the downstream SDK users.
  - They're not performance sensitive (turned off on validators, not part of canonical history) so:
    - They can contain a mix of `SCError`-like error codes, static messages, and also live / dynamic data, like the example above that shows the requested value as well as the limit. This allows far more meaningful diagnostics.
    - They can contain _multiple_ events, producing a trail of things-that-happened, including errors but also including general step-by-step debugging information. We can even vary the granularity of that, have core or the env dial up or down the debug level while it runs.

tsachiherman: @graydon the above, from my perspective, is beyond what we need. As long as we'll get these for all error cases, I think it would work. Unfortunately, consuming that data structure would require some more work on the downstream systems, so I wouldn't expect it to be fully supported for the testnet release.

ire-and-curses: I think this looks great for aiding with development and debugging - thanks!

graydon:
Closing this since I believe we're not doing it.

PR #147, opened 2023-09-07
Next into curr

sisuresh:
~~Marked as draft because the history includes every commit in next for some reason...~~ Edit: I don't think we have another choice with the merge because there's no common history between the two branches.

I had to merge with `git merge --allow-unrelated-histories upstream/next` to get this to work.


.github/workflows/check.yml
```diff
@@ -1,15 +1,30 @@
...
 name: CI
 
+# Controls when the workflow will run
 on:
+  # Triggers the workflow on push or pull request events for main, curr or next.
   push:
-    branches: [ "curr", "next" ]
+    branches: [ "main", "curr", "next" ]
   pull_request:
+    branches: [ "main", "curr", "next" ]
```
graydon:
```suggestion
    branches: [ "curr", "next" ]
  pull_request:
    branches: [ "curr", "next" ]
```

graydon (approved):

.github/workflows/check.yml
graydon: I _think_ this will be necessary to make CI work, there is no "main" branch

PR #155, opened 2023-10-09
Change the semantics of Soroban `refundableFee` to be `resourceFee` cvering all the resources

dmkozh:
Change the semantics of Soroban `refundableFee` to be `resourceFee` cvering all the resources.

The motivation is to allow stable prioritization of transactions independent of the network configuration (the inclusion fee is always constant).

### DIFF
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -821,8 +821,16 @@ struct SorobanTransactionData
 {
     ExtensionPoint ext;
     SorobanResources resources;
-    // Portion of transaction `fee` allocated to refundable fees.
-    int64 refundableFee;
+    // Portion of transaction `fee` allocated to the Soroban resource fees.
+    // The fraction of `resourceFee` corresponding to `resources` specified 
+    // above is *not* refundable (i.e. fees for instructions, ledger I/O), as
+    // well as fees for the transaction size.
+    // The remaining part of the fee is refundable and the charged value is
+    // based on the actual consumption of refundable resources (events, ledger
+    // rent bumps).
+    // The `inclusionFee` used for prioritization of the transaction is defined
+    // as `tx.fee - resourceFee`.
+    int64 resourceFee;
 };
 
 // TransactionV0 is a transaction with the AccountID discriminant stripped off,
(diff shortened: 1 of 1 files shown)


Stellar-transaction.x
```diff
@@ -821,8 +821,16 @@ struct SorobanTransactionData
...
+    // Portion of transaction `fee` allocated to the Soroban resource fees.
+    // The fraction of `resourceFee` corresponding to `resources` specified 
+    // above is *not* refundable (i.e. fees for instructions, ledger I/O), as
+    // well as fees for the transaction size.
+    // The remaining part of the fee is refundable and the charged value is
+    // based on the actual consumption of refundable resources (events, ledger
+    // rent bumps).
+    // The `inclusionFee` used for prioritization of the transaction is defined
+    // as `tx.fee - resourceFee`.
+    int64 resourceFee;
```
leighmcculloch: I haven't been following the history of why we're making this change, but even after reading the comment, it is pretty hard to understand how this field is used and the affect it has. I think it will be difficult to explain to others.

dmkozh: The comment doesn't tell 'why' - the 'why' is in the PR description and basically the reason is that we want to have stable tx prioritization logic (the alternative to that is quite exhaustive core refactoring to avoid some annoying UB cases). As for the field semantics, the suggestions on how to make this more clear are welcome. Previously, this was just the refundable part of the resource fees (for stuff like events/rent bumps). Now it includes all the soroban resource fees which, IMHO, is relatively more intuitive/simple.

leighmcculloch (approved):

Stellar-transaction.x
```diff
@@ -821,8 +821,16 @@ struct SorobanTransactionData
 {
     ExtensionPoint ext;
     SorobanResources resources;
-    // Portion of transaction `fee` allocated to refundable fees.
-    int64 refundableFee;
+    // Portion of transaction `fee` allocated to the Soroban resource fees.
```
leighmcculloch:
```suggestion
    // Amount of the transaction `fee` allocated to the Soroban resource fees.
```

PR #157, opened 2023-10-09
Rename state expiration to state archival

SirTyson:
Renames State Expiration terms as discussed [here](https://discord.com/channels/897514728459468821/1159192183870607452).

### DIFF
--- a/Stellar-ledger-entries.x
+++ b/Stellar-ledger-entries.x
@@ -101,7 +101,7 @@ enum LedgerEntryType
     CONTRACT_DATA = 6,
     CONTRACT_CODE = 7,
     CONFIG_SETTING = 8,
-    EXPIRATION = 9
+    TTL = 9
 };
 
 struct Signer
@@ -515,10 +515,10 @@ struct ContractCodeEntry {
     opaque code<>;
 };
 
-struct ExpirationEntry {
-    // Hash of the LedgerKey that is associated with this ExpirationEntry
+struct TTLEntry {
+    // Hash of the LedgerKey that is associated with this TTLEntry
     Hash keyHash;
-    uint32 expirationLedgerSeq;
+    uint32 liveUntilLedgerSeq;
 };
 
 struct LedgerEntryExtensionV1
@@ -557,8 +557,8 @@ struct LedgerEntry
         ContractCodeEntry contractCode;
     case CONFIG_SETTING:
         ConfigSettingEntry configSetting;
-    case EXPIRATION:
-        ExpirationEntry expiration;
+    case TTL:
+        TTLEntry ttl;
     }
     data;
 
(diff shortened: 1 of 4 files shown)


Stellar-ledger-entries.x
```diff
@@ -515,10 +515,10 @@ struct ContractCodeEntry {
     opaque code<>;
 };
 
-struct ExpirationEntry {
-    // Hash of the LedgerKey that is associated with this ExpirationEntry
+struct TTLEntry {
+    // Hash of the LedgerKey that is associated with this TTLEntry
     Hash keyHash;
-    uint32 expirationLedgerSeq;
+    uint32 liveUntilLedgerSeq;
```
leighmcculloch, line 521: Is this field inclusive or exclusive?

SirTyson: Inclusive, such that an entry e is live iff `e.liveUntilLedgerSeq >= currentLedgerSeq`

dmkozh (approved):

Stellar-ledger-entries.x
leighmcculloch: Then I think the name is misleading. I think it's reasonable to interpret until to mean exclusive. It's alive until the given ledger seq where it will now no longer be live.

anupsdf: > Then I think the name is misleading. I think it's reasonable to interpret until to mean exclusive. It's alive until the given ledger seq where it will now no longer be live.

Until means up to that ledger, right? We can also consider `liveTilLedgerSeq`.

SirTyson: I think the issue with this name is that we have to name is such that "this entry lives until X." Previously we used the language "expires on". With the new naming convention, we can't use the "expires on" phrasing because there is not a unified verb for what happens when a TEMPORARY and PERSISTENT entry "expire." When they are no longer live, TEMPORARY entries get deleted, while PERSISTENT entries get archived. We could use something like "deathLedger", but I'd like to avoid terminology suggesting entries are deleted. Given that we no longer have a unifying verb (i.e. expired), I think using the live until phrasing is appropriate. This is mostly an implementation detail, I expect most devs to interface with this information via and entries TTL that is derived from this field. I don't thing `liveTilLedgerSeq` helps solve Leigh's issue since it's the same word. Do you have any suggestions that indicate this field to be inclusive @leighmcculloch?

Stellar-transaction.x
```diff
@@ -585,19 +585,19 @@ struct InvokeHostFunctionOp
...
     Threshold: med
-    Result: BumpFootprintExpirationResult
+    Result: ExtendFootprintTTLResult
 */
-struct BumpFootprintExpirationOp
+struct ExtendFootprintTTLOp
 {
     ExtensionPoint ext;
-    uint32 ledgersToExpire;
+    uint32 extendTo;
```
2opremio, line 597: Does this mean that the field semantics changes to an absolute value and not relative?

2opremio: i.e. is this just a rename?

PR #163, opened 2023-10-11
Use liveUntilLedgerSeq in auth

sisuresh:


dmkozh:
I've noticed this in the previous PR, and I don't think we need the rename. Signature actually expires, so I think it's useful to distinguish it from the new state archival naming.

PR #173, opened 2024-03-18
Add Soroban write fee to LCM.

dmkozh:
Write fee is pretty hard/error-prone to compute manually, so it's easier to just export it. This also automatically supports users in case if the underlying write fee math changes.

This is necessary work as a part of https://github.com/stellar/stellar-core/issues/4244.

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -484,11 +484,22 @@ struct LedgerCloseMetaV0
     SCPHistoryEntry scpInfo<>;
 };
 
+struct LedgerCloseMetaExtV1
+{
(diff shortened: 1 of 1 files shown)


sisuresh:
Duplicate of https://github.com/stellar/stellar-xdr/pull/174

dmkozh:
This is not a duplicate, we still have 2 XDR branches, so we need to have the change in both.

sisuresh:
> This is not a duplicate, we still have 2 XDR branches, so we need to have the change in both.

Ah I missed the branch. Sorry about that.

sisuresh (approved):

PR #174, opened 2024-03-18
Add Soroban write fee to LCM.

dmkozh:
Write fee is pretty hard/error-prone to compute manually, so it's easier to just export it. This also automatically supports users in case if the underlying write fee math changes.

This is necessary work as a part of https://github.com/stellar/stellar-core/issues/4244.

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -484,11 +484,22 @@ struct LedgerCloseMetaV0
     SCPHistoryEntry scpInfo<>;
 };
 
+struct LedgerCloseMetaExtV1
+{
+    int64 sorobanFeeWrite1KB;
+};
+
+union LedgerCloseMetaExt switch (int v)
+{
+case 0:
+    void;
+case 1:
+    LedgerCloseMetaExtV1 v1;
+};
+
 struct LedgerCloseMetaV1
 {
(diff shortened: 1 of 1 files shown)


Stellar-ledger.x
```diff
@@ -484,11 +484,22 @@ struct LedgerCloseMetaV0
     SCPHistoryEntry scpInfo<>;
 };
 
+struct LedgerCloseMetaExtV1
+{
```
sisuresh, line 488: There should be an `ExtensionPoint` here to allow for future extensions.

dmkozh: I'm not sure whether this is the right way to go - wouldn't v2 extension be better? Working with nested extensions is not fun (from my recent experience of populating account ext v3 in Rust).

sisuresh: We mostly use nested extensions at the moment, so I was thinking it'd make sense for consistency. It also lets you adapt by only adding new logic to handle the nested extension because they parent struct stays the same.

sisuresh: There's only one field here though so a v2 extension is fine if that's the way you think we should go.

dmkozh: Ok, I've added the extension point, we can decide later whether to use it.

sisuresh (approved):

PR #178, opened 2024-03-25
Add cost types for secp256r1

jayz22:
XDR changes for https://github.com/stellar/rs-soroban-env/issues/807

### DIFF
--- a/Stellar-contract-config-setting.x
+++ b/Stellar-contract-config-setting.x
@@ -124,8 +124,9 @@ enum ContractCostType {
     InvokeVmFunction = 13,
     // Cost of computing a keccak256 hash from bytes.
     ComputeKeccak256Hash = 14,
-    // Cost of computing an ECDSA secp256k1 signature from bytes.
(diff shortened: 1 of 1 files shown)


Stellar-contract-config-setting.x
dmkozh, line 1: I *think* that this can also be merged into 'curr' XDR branch. This is just an enum, so there shouldn't be a way to somehow use this in 20.4 build.

dmkozh (approved):

Stellar-contract-config-setting.x
jayz22: Reasonable. But since this is designated to be used for the next protocol only, I think we can just put this in the `next` branch and think less about it. There are already other cost types (from Vm instantiation work) that are merged in `next` but not `curr`.

dmkozh: I think we should try to keep these branches in sync unless there is a change that we want to isolate from the current protocol builds. But since there is already a discrepancy, I guess it's ok to wait for a few more days before adding all the changes to curr.

PR #187, opened 2024-05-09
Add an XDR type for DiagnosticEvent<>

leighmcculloch:
### What
Add an XDR type DiagnosticEvents for DiagnosticEvent<>.

### Why
Stellar-core returns a variable length array of diagnostic events from it's submission endpoint. It'd be convenient to be able to decode it on the command line with the CLI. We could add the ability to decode arbitrary arrays and other types to the CLI, but that's much more effort and adding a type definition for a type that stellar-core does output is easier and maybe the "right" way to do this anyway.

Supersedes:
- https://github.com/stellar/rs-stellar-xdr/pull/366

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -400,6 +400,8 @@ struct DiagnosticEvent
     ContractEvent event;
 };
 
+typedef DiagnosticEvent DiagnosticEvents<>;
+
 struct SorobanTransactionMetaExtV1
 {
     ExtensionPoint ext;


leighmcculloch:
@MonsieurNicolas Does this need a protocol version? I don't think so. It's not used by anything explicitly, and implicitly the part of stellar-core that outputs this type (without using the type) is not part of the protocol anyway.

dmkozh (approved):

PR #195, opened 2024-07-19
Added Binary Fuse Filter XDR definitions

SirTyson:
Adds Binary Fuse Filter serialization struct to XDR. Note that this is internal to core implementation (for now) and should not break downstream or overlay.

### DIFF
--- a/Stellar-types.x
+++ b/Stellar-types.x
@@ -103,4 +103,35 @@ struct HmacSha256Mac
 {
     opaque mac[32];
 };
+
+struct ShortHashSeed
+{
+    opaque seed[16];
+};
+
+enum BinaryFuseFilterType
+{
+    BINARY_FUSE_FILTER_8_BIT = 0,
+    BINARY_FUSE_FILTER_16_BIT = 1,
+    BINARY_FUSE_FILTER_32_BIT = 2
+};
+
+struct SerializedBinaryFuseFilter
+{
+    BinaryFuseFilterType type;
+
+    // Seed used to hash input to filter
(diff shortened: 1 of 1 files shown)


dmkozh (approved):

Stellar-types.x
```diff
@@ -103,4 +103,35 @@ struct HmacSha256Mac
...
+    // Seed used for internal filter hash operations
+    ShortHashSeed filterSeed;
+    uint32 segmentLength;
+    uint32 segementLengthMask;
+    uint32 segmentCount;
+    uint32 segmentCountLength;
+    uint32 fingerprintLength; // Length in terms of element count, not bytes
+
+    // Array of uint8_t, uint16_t, or uint32_t depending on filter type
+    opaque fingerprints<>;
```
graydon, line 135: Sorry I didn't see this earlier -- you could make this a somewhat more robust definition if you make the 3 types of fingerprints arrays a union of 3 separate array types keyed by the filter type.

SirTyson: I thought about that, but the issue is there's no 16 bit XDR type. I could do an opaque vector for 8 bit filters, and an unsigned int vector for 32 bit, but I'd either have to define a 16 bit type or use an opaque vector and bit pack it manually for 16 bit values. Since it seemed like I had to manually pack bits for the 16 bit filter anyway, it seemed simpler just to have an opaque array here and manage the bit packing for all types instead of having different logic for each type.

graydon: Ah right. Fair point.

PR #200, opened 2024-07-30
Hot archive

SirTyson:
Adds `Hot Archive` types to vNext.

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -157,6 +157,13 @@ struct ConfigUpgradeSet {
     ConfigSettingEntry updatedEntry<>;
 };
 
+enum BucketListType
+{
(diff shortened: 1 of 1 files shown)


Stellar-ledger.x
```diff
@@ -168,6 +175,16 @@ enum BucketEntryType
     INITENTRY = 2 // At-and-after protocol 11: only created.
 };
 
+enum HotArchiveBucketEntryType
+{
+    HA_METAENTRY = -1, // Bucket metadata, should come first.
```
dmkozh: nit: shall we call this `HOST_ARCHIVE_...`? I think that's both more readable and safer (given that IIRC XDR enums are compiled into C enums).

dmkozh (approved):

PR #211, opened 2024-08-27
Add bls12-381 cost types

jayz22:


anupsdf:
Instead of separate G1 and G2 cost types should we have just one? Then we can call something like Bls12381GroupAdd, Bls12381GroupMul etc. 

Looking at the equation seems like Fq2 is constant,
```
G1 curve equation: y^2 = x^3 + 4
G2 curve equation: y^2 = x^3 + Fq2(4, 4)
```

jayz22:
> Instead of separate G1 and G2 cost types should we have just one? Then we can call something like Bls12381GroupAdd, Bls12381GroupMul etc.
> 
> Looking at the equation seems like Fq2 is constant,
> 
> ```
> G1 curve equation: y^2 = x^3 + 4
> G2 curve equation: y^2 = x^3 + Fq2(4, 4)
> ```

No if the groups are different, the costs are different.

anupsdf (approved):

PR #220, opened 2024-09-24
Add restoration and state archival meta

SirTyson:


Stellar-ledger-entries.x
```diff
@@ -795,3 +794,4 @@ case COLD_ARCHIVE_BOUNDARY_LEAF:
 case COLD_ARCHIVE_HASH:
     ColdArchiveHashEntry hashEntry;
 };
+}
```
leighmcculloch: Undo eofnl flipping:
```suggestion
}

```

Stellar-ledger.x
```diff
@@ -305,6 +306,8 @@ case LEDGER_ENTRY_REMOVED:
     LedgerKey removed;
 case LEDGER_ENTRY_STATE:
     LedgerEntry state;
+case LEDGER_ENTRY_RESTORE:
+    LedgerEntry restored;
```
leighmcculloch: When ledger entries get archived, which `LedgerEntryChange` arm occurs at that moment? Is it `LEDGER_ENTRY_REMOVED`? Or a new one not added yet?

leighmcculloch (commented):
One question.

Stellar-ledger.x
SirTyson: Currently there is no meta emitted when an entry is archived. Restores currently fire a `LEDGER_ENTRY_UPDATED` for both the TTL and restored entry. This will change to both the TTL and entry being of type `LEDGER_ENTRY_RESTORE`

SirTyson: Currently there is no meta emitted when an entry is archived. Restores currently fire a LEDGER_ENTRY_UPDATED for both the TTL and restored entry. This will change to both the TTL and entry being of type LEDGER_ENTRY_RESTORE

leighmcculloch: The lack of symmetry, and inability to identify in the meta archived entries seems problematic. How does someone ingesting discover archived entries?

dmkozh: > How does someone ingesting discover archived entries?

There should not be a need to discover archived entries. The latest state of entry will tell when it expires. There is no need to couple this with an entry being 'physically' archived by Core.

SirTyson:
Closing, this is supposed to be in `next`, not `curr`. See: https://github.com/stellar/stellar-xdr/pull/221

PR #221, opened 2024-09-25
Add restoration and state archival meta

SirTyson:
Same PR as before, I had mistakenly opened it against `curr` instead of `next`.

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -300,10 +300,11 @@ case 0:
 
 enum LedgerEntryChangeType
 {
-    LEDGER_ENTRY_CREATED = 0, // entry was added to the ledger
-    LEDGER_ENTRY_UPDATED = 1, // entry was modified in the ledger
-    LEDGER_ENTRY_REMOVED = 2, // entry was removed from the ledger
-    LEDGER_ENTRY_STATE = 3    // value of the entry
+    LEDGER_ENTRY_CREATED  = 0, // entry was added to the ledger
+    LEDGER_ENTRY_UPDATED  = 1, // entry was modified in the ledger
+    LEDGER_ENTRY_REMOVED  = 2, // entry was removed from the ledger
+    LEDGER_ENTRY_STATE    = 3, // value of the entry
+    LEDGER_ENTRY_RESTORED = 4  // archived entry was restored in the ledger
 };
 
 union LedgerEntryChange switch (LedgerEntryChangeType type)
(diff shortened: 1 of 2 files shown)


Stellar-ledger.x
```diff
@@ -506,12 +509,26 @@ struct LedgerCloseMetaExtV1
     int64 sorobanFeeWrite1KB;
 };
 
+struct LedgerCloseMetaExtV2
```
dmkozh, line 512: Shall we make this extensible? Extension nesting is not super readable, but I think it's a better option for the flag-protected meta changes in minor releases.

Stellar-ledger.x
```diff
@@ -506,12 +509,26 @@ struct LedgerCloseMetaExtV1
...
+{
+    int64 sorobanFeeWrite1KB;
+
+    uint32 currentArchivalEpoch;
+
+    // The last epoch currently stored by validators
+    // Any entry restored from an epoch older than this will
+    // require a proof.
+    uint32 lastArchivalEpochPersisted;
```
dmkozh, line 522: How is this supposed to be used? I thought Core will tell when a proof is necessary.

SirTyson: I don't think downstream will currently use it, but my goal here was for meta to have sufficient information such that somebody would be able to construct proofs without captive-core. It would take a lot of work and I don't think it's likely someone will actually do it, but with the curerntArchivalEpoch, the lastArchivalEpochPersisted, and the eviction meta, a downstream system would be able to maintain all merkle trees and construct restoration proofs correctly without using captive-core BucketListDB endpoints.

Stellar-transaction.x
```diff
@@ -1876,7 +1876,8 @@ enum InvokeHostFunctionResultCode
     INVOKE_HOST_FUNCTION_TRAPPED = -2,
     INVOKE_HOST_FUNCTION_RESOURCE_LIMIT_EXCEEDED = -3,
     INVOKE_HOST_FUNCTION_ENTRY_ARCHIVED = -4,
-    INVOKE_HOST_FUNCTION_INSUFFICIENT_REFUNDABLE_FEE = -5
+    INVOKE_HOST_FUNCTION_INSUFFICIENT_REFUNDABLE_FEE = -5,
+    INVOKE_HOST_FUNCTION_INVALID_PROOF = -6
```
dmkozh: nit: `INVALID_RESTORATION_PROOF` or something along these lines? Just 'proof' is ambiguous.

dmkozh (approved):

PR #224, opened 2024-11-05
minor cleanup, curr

marta-lokhova:


Stellar-overlay.x
```diff
@@ -94,7 +94,6 @@ enum MessageType
     AUTH = 2,
     DONT_HAVE = 3,
 
-    GET_PEERS = 4, // gets a list of peers this guy knows about
```
leighmcculloch, line 97: Not sure if we typically do this, but may be some value to leaving a comment indicating that `4` shouldn't be reused. Or maybe there's no harm in reuse after a certain point so there's little value in adding it?

leighmcculloch (approved):
I don't have a lot of context on this change, but looks like clean up of a value that will be no longer used. One comment, but lgtm regardless.

Stellar-overlay.x
marta-lokhova: Good idea! I left a comment that GET_PEERS is deprecated. I did not mention not re-using it, since at some point it would be safe to re-use it (once overlay drops support for GET_PEERS completely, for example)

anupsdf (approved):

PR #232, opened 2025-02-12
Cap 62

SirTyson:
Cleans up some outdated structs and renames evictionMeta field for CAP 62.

### DIFF
--- a/Stellar-ledger-entries.x
+++ b/Stellar-ledger-entries.x
@@ -682,8 +682,7 @@ enum EnvelopeType
 enum BucketListType
 {
     LIVE = 0,
-    HOT_ARCHIVE = 1,
-    COLD_ARCHIVE = 2
+    HOT_ARCHIVE = 1
 };
 
 /* Entries used to define the bucket list */
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -823,56 +823,6 @@ struct LedgerFootprint
     LedgerKey readWrite<>;
 };
 
-enum ArchivalProofType
-{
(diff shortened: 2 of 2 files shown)


Stellar-ledger.x
```diff
@@ -528,12 +528,11 @@ struct LedgerCloseMetaV1
     // systems calculating storage fees correctly.
     uint64 totalByteSizeOfBucketList;
 
-    // Temp keys that are being evicted at this ledger.
-    LedgerKey evictedTemporaryLedgerKeys<>;
+    // CONTRACT_CODE/CONTRACT_DATA and their associated TTL keys that are being evicted.
+    LedgerKey evictedKeys<>;
```
dmkozh: I think we should postpone this change in the `curr` branch until after the release.

Stellar-ledger-entries.x
```diff
@@ -682,8 +682,7 @@ enum EnvelopeType
 enum BucketListType
 {
     LIVE = 0,
-    HOT_ARCHIVE = 1,
-    COLD_ARCHIVE = 2
+    HOT_ARCHIVE = 1
```
dmkozh, line 685: Are we sure this is safe to do in `curr` xdr?

SirTyson: I think it's fine. We've had these structs in curr for a while now, I'm just removing the unused ones. It's internal to core so I don't think there would be any issue, and if we remove the HOT_ARCHIVE stuff we'd need to do a bunch of ifdefs in core before the release.

dmkozh (approved):

PR #233, opened 2025-02-12
Cap 66 next

SirTyson:
Removes proofs and cold archive structs from next.

Additionally, adds `archivedSorobanEntries` for CAP 66. While CAP 66 has other XDR changes as well, these are all renames that will be done after the next core release so that we can do them in both next and curr at the same time to avoid messy ifdefs.

### DIFF
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -848,56 +848,6 @@ struct LedgerFootprint
     LedgerKey readWrite<>;
 };
 
-enum ArchivalProofType
-{
-    EXISTENCE = 0,
-    NONEXISTENCE = 1
-};
-
-struct ArchivalProofNode
-{
-    uint32 index;
-    Hash hash;
-};
-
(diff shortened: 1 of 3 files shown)


Stellar-transaction.x
```diff
@@ -888,6 +838,14 @@ struct SorobanResources
     uint32 writeBytes;
 };
 
+struct SorobanResourcesExtV0
+{
+    // Vector of uint8_t indices representing what Soroban
+    // entries in the footprint are archived, based on the
+    // order of keys provided in the readWrite footprint.
+    opaque<> archivedSorobanEntries;
```
dmkozh: We've had a bit of discussion this on the CAP PR before; I think that the safest option here would be to use just the vector of `uint32`s. I'm worried that using the byte vector may trip up downstream systems, e.g. mess up tx building logic in JS client, or mess up the rendering in any XDR viewer (lab, chain explorer etc.)

dmkozh (approved):

PR #241, opened 2025-02-27
XDR changes for the muxed address support sketch - with memo muxed account support

dmkozh:


Stellar-contract-spec.x
```diff
@@ -105,6 +106,8 @@ case SC_SPEC_TYPE_STRING:
 case SC_SPEC_TYPE_SYMBOL:
 case SC_SPEC_TYPE_ADDRESS:
     void;
+case SC_SPEC_TYPE_ADDRESS_V2:
+    bool supportMultiplexing;
```
leighmcculloch, line 110: I still think this should be a separate type.
```suggestion
case SC_SPEC_TYPE_ADDRESS_MUXED:
    void;
```
Why: So we create less valid options.

With AddressV2, we end up with the following possible cases:
- Address
- AddressV2, supportMultiplexing = false
- AddressV2, supportMultiplexing = true

With MuxedAddress, we end up with only two valid cases:
- Address
- MuxedAddress

As a secondary concern, which isn't really that important, the bool is 4-bytes.

leighmcculloch: As an optimisation, the SDK will probably only emit these versions with AddressV2:

- Address
- AddressV2, supportMultiplexing = true

dmkozh: Yeah, I think `MuxedAddress` is fine if we're fine with a bit trickier semantics. I just didn't have time to update the spike with this change; memo spike is unrelated to the contract spec.

PR #242, opened 2025-03-04
Remove support for old survey mechanism (curr)

bboston7:
This change removes all message types used in the old survey mechanism. This is required to support stellar/stellar-core#4332.

### DIFF
--- a/Stellar-overlay.x
+++ b/Stellar-overlay.x
@@ -112,6 +112,9 @@ enum MessageType
     // new messages
     HELLO = 13,
 
+    // SURVEY_REQUEST (14) removed and replaced by TIME_SLICED_SURVEY_REQUEST
+    // SURVEY_RESPONSE (15) removed and replaced by TIME_SLICED_SURVEY_RESPONSE
+
     SEND_MORE = 16,
     SEND_MORE_EXTENDED = 20,
 
--- a/Stellar-overlay.x
+++ b/Stellar-overlay.x
@@ -112,9 +112,6 @@ enum MessageType
     // new messages
     HELLO = 13,
 
-    SURVEY_REQUEST = 14,
-    SURVEY_RESPONSE = 15,
-
     SEND_MORE = 16,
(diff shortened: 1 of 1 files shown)


Stellar-overlay.x
```diff
@@ -310,10 +269,6 @@ struct TopologyResponseBodyV2
 
 union SurveyResponseBody switch (SurveyMessageResponseType type)
 {
-case SURVEY_TOPOLOGY_RESPONSE_V0:
-    TopologyResponseBodyV0 topologyResponseBodyV0;
-case SURVEY_TOPOLOGY_RESPONSE_V1:
-    TopologyResponseBodyV1 topologyResponseBodyV1;
 case SURVEY_TOPOLOGY_RESPONSE_V2:
```
anupsdf, line 275: This now leaves the `union` with a single arm which is fine since this leaves us with the future flexibility to add another survey.

Stellar-overlay.x
```diff
@@ -112,9 +112,6 @@ enum MessageType
     // new messages
     HELLO = 13,
 
-    SURVEY_REQUEST = 14,
```
marta-lokhova, line 115: Might be helpful to leave a short comment for posterity that these two message types were deprecated? (Similar to GET_PEERS)

marta-lokhova (approved):

Stellar-overlay.x
bboston7: Done!

Stellar-overlay.x
bboston7: Yeah, I went back and forth on this, but ultimately decided to leave this union in to make expanding the survey easier in the future.

marta-lokhova (approved):

marta-lokhova:
Both PRs LGTM, however I don't have write access so can't merge these. @anupsdf could you help?

anupsdf (approved):

PR #245, opened 2025-03-06
Add TransactionResultMetaV1 and LedgerCloseMetaV2

sisuresh:
What it would look like if we were to add a `postTxApplyProcessing` portion of meta. This is the xdr change for one of the solutions for dealing with fee refunds in parallel soroban.

### DIFF
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -506,7 +506,7 @@ struct TransactionResultMetaV1
     LedgerEntryChanges feeProcessing;
     TransactionMeta txApplyProcessing;
 
-    LedgerEntryChanges postTxApplyProcessing;
+    LedgerEntryChanges postTxApplyFeeProcessing;
 };
 
 // this represents a single upgrade that was performed as part of a ledger
--- a/Stellar-ledger.x
+++ b/Stellar-ledger.x
@@ -488,11 +488,25 @@ case 3:
 // This struct groups together changes on a per transaction basis
 // note however that fees and transaction application are done in separate
 // phases
-struct TransactionResultMeta
+struct TransactionResultMetaV0
 {
     TransactionResultPair result;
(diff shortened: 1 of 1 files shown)


Stellar-ledger.x
```diff
@@ -488,13 +488,27 @@ case 3:
...
+// phases
+struct TransactionResultMetaV1
+{
+    ExtensionPoint ext;
+
+    TransactionResultPair result;
+    LedgerEntryChanges feeProcessing;
+    TransactionMeta txApplyProcessing;
+
+    LedgerEntryChanges postTxApplyProcessing;
```
dmkozh: nit: This should probably be called `postTxApplyFeeProcessing` to reflect the fact that only fee changes are expected here.

sisuresh: I intentionally made it more generic for flexibility, but I didn't have any specific use cases in mind. I'm fine with `postTxApplyFeeProcessing`.

dmkozh: I just think that if we want something more generic, we shouldn't go with raw ledger entry changes, as changes would become hard to interpret.

sisuresh: Yeah that's valid. We can extend `TransactionResultMetaV1` if we need to.

Stellar-ledger.x
```diff
@@ -488,13 +488,27 @@ case 3:
 // This struct groups together changes on a per transaction basis
 // note however that fees and transaction application are done in separate
 // phases
-struct TransactionResultMeta
+struct TransactionResultMetaV0
```
dmkozh: Is this change actually necessary? I think leaving this as is would decrease the amount of the necessary downstream changes.

sisuresh: Removed

dmkozh (approved):

PR #278, opened 2025-05-14
Rename totalByteSizeOfLiveSorobanState

SirTyson:


Stellar-ledger.x
```diff
@@ -589,9 +589,9 @@ struct LedgerCloseMetaV1
     // other misc information attached to the ledger close
     SCPHistoryEntry scpInfo<>;
 
-    // Size in bytes of BucketList, to support downstream
+    // Size in bytes of live Soroban state, to support downstream
     // systems calculating storage fees correctly.
-    uint64 totalByteSizeOfBucketList;
+    uint64 totalByteSizeOfLiveSorobanState;
```
dmkozh: This has to be rebased and the change should also go to meta v2.

SirTyson: Hmm the git diff display was kinda funky, we already have the rename in v1. Rebased, should be fine now.

Stellar-ledger.x
```diff
@@ -658,9 +658,9 @@ struct LedgerCloseMetaV2
...
 
     // Temp keys and all TTL keys that are being evicted at this ledger.
     // Note that this can contain TTL keys for both persistent and temporary
```
dmkozh: What about `evictedTemporaryLedgerKeys` and `evictedPersistentLedgerEntries`? We should rename one and remove the other, right?

dmkozh (approved):

PR #282, opened 2025-08-05
Add padding transaction op for testing

drebelsky:
In core loadgen, it would be useful if we had an operator we could use to pad out the size of a transaction frame. See also https://github.com/stellar/stellar-core/pull/4866.

### DIFF
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -64,7 +64,8 @@ enum OperationType
     LIQUIDITY_POOL_WITHDRAW = 23,
     INVOKE_HOST_FUNCTION = 24,
     EXTEND_FOOTPRINT_TTL = 25,
-    RESTORE_FOOTPRINT = 26
+    RESTORE_FOOTPRINT = 26,
+    PAD_FOR_TESTING = -1
 };
 
 /* CreateAccount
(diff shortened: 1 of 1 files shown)


Stellar-transaction.x
```diff
@@ -695,6 +707,8 @@ struct Operation
         ExtendFootprintTTLOp extendFootprintTTLOp;
     case RESTORE_FOOTPRINT:
         RestoreFootprintOp restoreFootprintOp;
+    case PAD_FOR_TESTING:
+        PadForTestingOp padForTestingOp;
```
leighmcculloch, line 711: Is there a way to be able to do this without introducing a new operation? Maybe a development feature of xdrpp to be able to specify during codegen a union type that gets additional padding arm for testing? Presumably this would be a useful thin to be able to do to any type.

Adding it to the XDR definition itself will cause all downstream systems and integrators to see `PadForTesting` as a valid Stellar operation they need to handle when decoding operations (e.g. Rust decoders will be forced to handle the arm), or display in UIs of development tools, and will be included in things like the JSON schema, etc.

drebelsky: Thanks for flagging this: I'll investigate alternative solutions.

drebelsky:
We found an alternative method for core—it's probably worth fixing the typo in a different PR, though.

PR #296, opened 2026-03-03
XDR changes for CAP-77.

dmkozh:
This introduces the configuration settings for modifying the 'frozen' ledger keys via validator vote, settings for allowing certain transactions to bypass the freeze, and the error codes for transactions that access frozen keys.

CAP-77: https://github.com/stellar/stellar-protocol/blob/master/core/cap-0077.md

### DIFF
--- a/Stellar-transaction.x
+++ b/Stellar-transaction.x
@@ -1597,7 +1597,8 @@ enum ClaimClaimableBalanceResultCode
     CLAIM_CLAIMABLE_BALANCE_CANNOT_CLAIM = -2,
     CLAIM_CLAIMABLE_BALANCE_LINE_FULL = -3,
     CLAIM_CLAIMABLE_BALANCE_NO_TRUST = -4,
-    CLAIM_CLAIMABLE_BALANCE_NOT_AUTHORIZED = -5
+    CLAIM_CLAIMABLE_BALANCE_NOT_AUTHORIZED = -5,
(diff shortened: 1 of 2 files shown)


Stellar-transaction.x
```diff
@@ -1597,7 +1597,8 @@ enum ClaimClaimableBalanceResultCode
     CLAIM_CLAIMABLE_BALANCE_CANNOT_CLAIM = -2,
     CLAIM_CLAIMABLE_BALANCE_LINE_FULL = -3,
     CLAIM_CLAIMABLE_BALANCE_NO_TRUST = -4,
-    CLAIM_CLAIMABLE_BALANCE_NOT_AUTHORIZED = -5
+    CLAIM_CLAIMABLE_BALANCE_NOT_AUTHORIZED = -5,
+    CLAIM_CLAIMABLE_BALANCE_TRUSTLINE_FROZEN = -6
```
sisuresh, line 1601: Is the create claimable balance error missing?

dmkozh: No, we should be able to tell if a balance is created from a frozen account/trustline. The per-operation error codes are only for the cases where we can't tell which account/trustline is going to be accessed by simply looking at tx.

dmkozh: To be clear, for validation we have a blanket tx-level error code.

sisuresh:
We'll wait until the core PR is ready before merging this PR.

sisuresh (approved):

PR #315, opened 2026-08-13
Exclude SCV_EXECUTABLE_TAG from the spec check

leighmcculloch:
### What

Add `EXECUTABLE_TAG` to the excluded variants in `.scripts/check-scspectype-superset.ts`, with a comment recording why it is excluded despite being passable across a contract function boundary.

### Why

The check has failed on `main` since it merged in #314, because `SCV_EXECUTABLE_TAG` from CAP-85 has no `SC_SPEC_TYPE_EXECUTABLE_TAG` — the SDK deliberately exposes the executable tag everywhere as a String for a better and more consistent UX, treating `SCV_EXECUTABLE_TAG` as a storage implementation detail, so no spec type is planned.

### DIFF
--- a/.scripts/check-scspectype-superset.ts
+++ b/.scripts/check-scspectype-superset.ts
@@ -4,12 +4,17 @@
 
 import { readFileSync } from "node:fs";
 
-// SCValType variants that cannot cross a contract function boundary, and so
-// intentionally have no SCSpecType counterpart.
+// SCValType variants that intentionally have no SCSpecType counterpart.
 const EXCLUDED = [
+  // Internal values that cannot cross a contract function boundary.
   "CONTRACT_INSTANCE",
   "LEDGER_KEY_CONTRACT_INSTANCE",
   "LEDGER_KEY_NONCE",
+  // Technically can be passed across a contract function boundary, but the SDK
+  // does not support it and does not plan to. For a better and more consistent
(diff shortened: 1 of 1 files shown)


.scripts/check-scspectype-superset.ts
```diff
@@ -4,12 +4,17 @@
...
 
-// SCValType variants that cannot cross a contract function boundary, and so
-// intentionally have no SCSpecType counterpart.
+// SCValType variants that intentionally have no SCSpecType counterpart.
 const EXCLUDED = [
+  // Internal values that cannot cross a contract function boundary.
   "CONTRACT_INSTANCE",
   "LEDGER_KEY_CONTRACT_INSTANCE",
   "LEDGER_KEY_NONCE",
+  // Technically can be passed across a contract function boundary, but the SDK
```
dmkozh, line 13: I would postpone this PR until we're 100% sure about the SDK approach.

leighmcculloch: We can wait. 👍 

But I think regardless of whether the sdk structures the logic as its own vertical or just as storage the sdk shouldnt encourage its use at boundaries since it's a marker intended just for storage.

dmkozh: The SDK change looks good to me, so this LGTM as well.

dmkozh (approved):

